PR #11, opened 2014-08-04
Support `else:` in conditional declarations

sinkuu:
```d
version (A) {
	void foo() {}
} else:
void bar() {}
void baz() {}
```
Else branches in conditional declarations can have multiple declarations. Libdparse cannot parse `else:`, and the AST node of `ConditionalDeclaration` is defined as
```d
...
Declaration[] trueDeclarations;
Declaration falseDeclaration;
...
```

So I fixed the parser, and changed that node definition to:
```d
...
Declaration[] trueDeclarations;
Declaration[] falseDeclarations;
...
```

### DIFF
--- a/src/std/d/ast.d
+++ b/src/std/d/ast.d
@@ -1060,11 +1060,11 @@ final class ConditionalDeclaration : ASTNode
 public:
     override void accept(ASTVisitor visitor) const
     {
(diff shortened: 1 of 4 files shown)


sinkuu:
It's likely that hardly anyone uses this syntax. I'll close.

Issue #41, opened 2015-03-09
FR: StaticConstructor and StaticDestructor line and column


BBasile:
this is a minor feature request: line and column info in the ASTNode `StaticConstructor`  and `StaticDestructor`.

BBasile:
feature implemented in #55

PR #55, opened 2015-05-20
Handling of .line, .column and .location in std.d.parse.simpleParse()

BBasile:
This modifications are proposed because of  #41 

- it doesn't break any previous usage of the AST nodes  _StaticConstructor_ and _StaticDestructor_ because the previous member _location_ is not replaced (even if useless since now line and col are more usable).
- notice that there was a doubt about _how-to_ because it would also be possible to implement the feature specifically for the two items instead of globally, in simpleParse().

### DIFF
--- a/src/std/d/ast.d
+++ b/src/std/d/ast.d
@@ -2528,6 +2528,8 @@ public:
     }
     /** */ FunctionBody functionBody;
     /** */ size_t location;
+    /** */ size_t line;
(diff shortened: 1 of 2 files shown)


Hackerpilot:
The *location* field is used by DCD for locating these AST nodes by byte offset. Your approach of adding this feature to simpleParse makes sense.

Issue #61, opened 2015-09-15
Parser.messageFunction should be a delegate


BBasile:
I find the fact that `Parser.messageFunction` is a function and not a delegate not very practical. 

When a parser is dedicated to a particular class instance, there is no easy way to determine if an item in the errors stack is related to an instance or another (apart from the filename but testing a string for each message is not the best way to identify an instance).

At least a user parameter could be given to the parser and passed for each message so that it would be possible to get a particular class instance.

So either the type of the callback should be changed to

```D
void delegate(string, size_t, size_t, string, bool)
```

or

```D
void function(string, size_t, size_t, string, bool, void*)`
```

With the last pointer the user param passed back by the parser, and in this case `parseModule` declaration would be something like that:

```D
Module parseModule(const(Token)[] tokens, string fileName, CAllocator allocator = null, 
  void function(string, size_t, size_t, string, bool) messageFunction = null, 
  void*  userParam = null, uint* errorCount = null, uint* warningCount = null)
// or maybe userParam at the end so that existing code des not break.
```

so that the callback can do that:

```D
class Foo
{
    struct AstError{}
    AstError*[] errors; // no static !!
    final static void parserError(string fname, size_t line, size_t col, string msg, bool isErr, void* userParam)
    {
        with( cast(Foo) userParam) errors ~= new AstError(line, col, msg, isErr);
    }
}
```

PR #75, opened 2016-01-01
added a user parameter to the messageFunction

BBasile:
The feature was proposed a while ago in #61.


Geod24:
It would be simpler and safer to just make it a delegate instead of doing it the C way.

PR #76, opened 2016-01-03
Use experimental_allocator from dub

keywan-ghadami:
i removed experimental_allocator git submodule because it isn't needed anymore.
Also removed version from dub file because version can be set by tags in github.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -4,7 +4,7 @@
 	"targetType": "library",
 	"description": "Library for lexing and parsing D source code",
 	"license": "BSL-1.0",
-        "version": "0.3.0-alpha",
-	"importPaths": ["src", "experimental_allocator"],
-	"sourcePaths": ["src", "experimental_allocator"]
+	"dependencies":{
+		"experimental_allocator":"~>0.0.1",
+	}
 }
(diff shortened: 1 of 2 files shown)


BBasile:
Take care to GDC and LDC. You must verify that their own latest phobos includes experimental.allocators.

Geod24:
As @BBasile mentioned, that would break building libdparse with any other compiler than DMD older than 2.069.0, which is dated November 3 (2015). The current LDC alpha is currently using FE 2.068.2 while GDC is using an amended 2.066.1.
While it's a good goal for the future, it's probably a bit too early to do.

As a side note, I would suggest to take a look at how to squash commits (google will provide some useful articles) if you haven't already. In this case, since the 3 commits are related to the same thing, they should probably be squashed into a single one to make reading the history easier.

BBasile:
And also I think that Brian's fork contains a fix that's only been pushed to phobos stable branch [yesterday](https://issues.dlang.org/show_bug.cgi?id=15188), for the upcoming 2.070 rlz.

keywan-ghadami:
Thank you, so closings this for now...

keywan-ghadami:
I am going to squash the commits and add the experimental_allocator with dub

BBasile:
This PR is totally obsolete now. closing.

PR #79, opened 2016-01-08
Fix #70: Parse error on `else:` in ConditionalDeclaration

### LINKED ISSUE #70: Parse error on 'else:'
```D
module tango.core.memory;

version(D_Version2)
{
    public import core.memory;
}
else:
```

test1.d(7:5)[error]: Declaration expected

Tested with master and 0.2.1

mathias-lang-sociomantic:
Allow code such as `version (X) {} else:` as DMD does.
As a result, there can be multiple `falseDeclaration` in a `conditionalStatement`,
so the single member was deprecated in favor of an array.

Fix Hackerpilot/libdparse#70

Note: It makes libdparse inevitably triggers a few deprecation warnings.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -1151,11 +1151,14 @@ final class ConditionalDeclaration : ASTNode
 public:
     override void accept(ASTVisitor visitor) const
     {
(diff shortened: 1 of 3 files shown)


mathias-lang-sociomantic:
Fixed the test failures.

PR #92, opened 2016-04-13
fixup for #91, line and col were set for the token following the closed paren

BBasile:


src/dparse/parser.d
```diff
@@ -7294,8 +7294,8 @@ protected:
         mixin(tokenCheck!"(");
         mixin(tokenCheck!")");
         StackBuffer attributes;
-        node.line = current.line;
-        node.column = current.column;
+        node.line = tokens[index - 3].line;
```
Hackerpilot: Why not use `current.line` but move the assignment to the top of the function?

BBasile:
done.

PR #104, opened 2016-06-17
Cloning and removing NIMBY

cyisfor:
Since the parameter to visit is const, there needs to be a way to deep clone tree nodes, to make filtering or modification of an AST easier. If you just remove the const, then you can modify nodes in place, and move them around, but not duplicate them, or use them as templates. Also there's some custom complete reimplementing of normal class derivation that really doesn't need to be there. I removed that as a way of facilitating cloning, but it's a good idea in general not to make ugly, hacked facsimilies of elegant, builtin constructs that already exist in D.


Hackerpilot:
>  I removed that as a way of facilitating cloning, but it's a good idea in general not to make ugly, hacked facsimilies of elegant, builtin constructs that already exist in D.

I see that you believe the following things:

1. I don't know about the "elegant builtin constructs"
1. There is no reason that the code is written the way it is.

cyisfor:
@Hackerpilot I didn't mean to be insulting, sorry. I think we've all had times where there seemed like a better solution, but it just ended up reinventing the wheel. A final class for instance, with a list of Variant types it acts as, is functionally identical to a non-final class being derived from. But I admittedly couldn't figure out certain oddities such as how a list of declarations was itself considered a declaration, or how to invoke the functions for the least, and the most-derived type. I'm fairly sure that's my failing to understand D though, not some inherent virtue to your strategy that I cannot comprehend. Your code is a lot easier to comprehend than dmd.

Anyway I just thought doing this was better than just whining impotently about the cloning issue in an issue report. Just reject the code, or whatever. It's already not worth bothering with, considering how many days I spent beating my head against it.

BBasile:
About changing the inheritance list: Each AST node should match to a rule of the grammar. By defining each possible declaration as a Declaration, you put the luggage of the Declaration (i.e the members) in nodes that are not made for the same rule. That's not as nice as you thought even if with the OO principles it would have been logical to do so.

About cloning: it's true that the const parameters in visit make the AST not modifiable. Since lifparse is only used for analysis this has not been a problem before.

BBasile:
This is going nowhere. closing.

PR #117, opened 2016-11-22
Improve Postblit parsing and add location, line and column

BBasile:
By the way I have a question:

`ParsePostblit` has:
```D
        expect(tok!"this");
        expect(tok!"(");
        expect(tok!"this");
        expect(tok!")");
```

but the `ParsePostblit` caller does already the lookup:

```D
            if (startsWith(tok!"this", tok!"(", tok!"this", tok!")"))
                mixin(parseNodeQ!(`node.postblit`, `Postblit`));
```  

Could the 4 `expect` be replaced by `index += 4` ? A few branches could be avoided in this case.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -2396,6 +2396,9 @@ public:
     }
     /** */ FunctionBody functionBody;
     /** */ MemberFunctionAttribute[] memberFunctionAttributes;
+    /** */ size_t location;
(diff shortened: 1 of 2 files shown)


BBasile:
Never mind the question, it's clear that the four `expect` are not needed. IN no other places the parser expects more than one time.

Issue #106, opened 2016-07-07
False positive in IfCondition parser


BBasile:
``` D
if (const a = 0){}
```

gives the error

> Expected ( instead of =

PR #126, opened 2017-02-03
fix regression in parseIfStatement

BBasile:
In #106 I broke 

```d
 if (TypeCtor(Type) ident = stuff){}
```

by allowing

```d
 if (TypeCtor ident = stuff){}
```

It's well tested now.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -3258,8 +3258,9 @@ class Parser
             expect(tok!"=");
             mixin(parseNodeQ!(`node.expression`, `Expression`));
         }
-        else if (isTypeCtor(current.type))
+        else
         {
+            // consume for TypeCtors = identifier
             while (isTypeCtor(current.type))
             {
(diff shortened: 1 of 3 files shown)


Hackerpilot:
I should have tested this more before tagging another beta. This change breaks this code:

```d
void main()
{
	if ((a && b) || c)
	{
	}
}
```

BBasile:
It must be
```d
            if (currentIs(tok!"("))
                index--;
```

that must not be called if not typector is consumed.

BBasile:
Indeed:

```d
            // consume for TypeCtors = identifier
            if (isTypeCtor(current.type))
            {
                while (isTypeCtor(current.type))
                {
                    const prev = current.type;
                    advance();
                    if (current.type == prev)
                        warn("redundant type constructor");
                }
                // goes back for TypeCtor(Type) = identifier
                if (currentIs(tok!"("))
                    index--;
            }
```

works for all.

PR #127, opened 2017-02-03
fix other regression in parseIfStatement

BBasile:
An open paren could happen without a previous TypeCtor.


Hackerpilot:
I built a version of D-Scanner with this change and it seems to parse Phobos correctly (excluding traits and typecons), so I think this is fixed now.

BBasile:
The regression was on master since a while and it's only appeared tonight with this story of scope. Quite unexpected !

PR #130, opened 2017-02-22
Formatter: fix fields of primaryExpression

wilzbach:
Previously for example, the `intLiteral` was omitted:

in:  assert(float(0))
out: assert(float)


wilzbach:
Thanks a lot for the quick merge @Hackerpilot 

Would you be so kind to release a new tag for libdparse, s.t. we could use it on dlang.org?
Unfortunately without workarounds DUB doesn't play well with git SHAs

PR #136, opened 2017-05-07
update badge address

BBasile:


README.md
```diff
@@ -6,7 +6,7 @@ Library for lexing and parsing D source code
 Online documentation is available [here](https://hackerpilot.github.io/libdparse/)
```
wilzbach: I guess this would be then https://dlang-community.github.io/libdparse ?

BBasile: Yes. Since you there right now can you re-invite me to the org ?

PR #139, opened 2017-05-19
fix #133 - only visit non empty Token

### LINKED ISSUE #133: Token get visited even if empty
I had the following code:

```D
module pragmamsg;

pragma(msg, int);

void main () {}
```

Running a visitor on this `ModuleDeclaration` that had the following override: `public override void visit (const DParse.Token d) { assert(0); }`, I got it triggered.

The backtrace includes:
```
./submodules/libdparse/src/dparse/ast.d:2175 const(void function(dparse.ast.ASTVisitor)) dparse.ast.Module.accept [0x803524]
```

Which is [here](https://github.com/Hackerpilot/libdparse/blob/v0.7.0-beta.3/src/dparse/ast.d#L2175).
Since the only `Token` in `ModuleDeclaration` is `scriptLine`, I assume that's what get visited.
IMO that shouldn't be the case, as there is no `scriptLine` in said code.

So far I'm working around this by checking `assert(d is Token.init);`.

BBasile:


Geod24:
Could you add a test case ?

mathias-lang-sociomantic:
Thanks :)

PR #141, opened 2017-05-20
Test Windows builds with AppVeyor

wilzbach:
A first experiment with enabling a CI for windows. Ideally should be enabled for all repos.
AppVeyor supports cronjobs as well.
This could be made faster if we run the x86 and x86_64 tests in one run, but then the main overview table would be less informative (I am fine with both ways).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,121 @@
+platform: x64
+environment:
+ matrix:
+  - DC: dmd
+    DVersion: nightly
+    arch: x64
+  - DC: dmd
+    DVersion: nightly
+    arch: x86
+  - DC: dmd
+    DVersion: beta
+    arch: x64
+  - DC: dmd
+    DVersion: beta
+    arch: x86
+  - DC: dmd
+    DVersion: stable
+    arch: x64
+  - DC: dmd
+    DVersion: stable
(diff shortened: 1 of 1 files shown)


wilzbach:
Wohoo :tada:
My tests on my fork (to avoid spamming with rebases) have worked:

image

wilzbach:
So do we go with testing on Windows?
If wanted we can always execute the `build.bat` (if existent for a repo) as well.
@rikkimax this might be a good starting point to automatically generate the binary executables for Windows.
After the build has succeeded you could let AppVeyor execute a simple D script that builds all binaries, e.g.
```bash
dub fetch --version="~master" dfix && dub build dfix
dub fetch --version="~master" dfmt && dub build dfmt
dub fetch --version="~master" dscanner && dub build dscanner
dub fetch --version="~master" dcd && dub build --config=client dcd && dub build --config=server dcd
```
and then upload them to s3. For more user convenience, you could even bundle them in an installer

wilzbach:
Ping @ZombineDev @BBasile @Hackerpilot

BBasile:
I've already given my opinion in the community discussions. The plus would be when the build.bat can be tested. since they tend to be less maintained.

PR #142, opened 2017-06-13
Copy UnionDeclaration comment

BBasile:
many libdparse-based software will have to be updated !

- DCD
- D-Scanner
- harbored-mod


BBasile:
> Looks like unions are highly unpopular ;-)

Yes that's truly unbelievable that nobody's noticed the bug last 4 years.

PR #144, opened 2017-06-13
[Experiment] Enable CodeCov

wilzbach:
In theory that should be all that's needed to get CodeCov reports on the CI status report.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,13 @@
+# Documentation: https://github.com/codecov/support/wiki/codecov.yml
+codecov:
+coverage:
+  precision: 3
+  round: down
+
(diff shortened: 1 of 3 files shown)


wilzbach:
hmm it seems that the `.codecov.yml` file is only taken into account when its part of the target branch.

keywan-ghadami-oxid:
Hi, yes it seams it is needed on the target branch too, to be able to compare results. But also it seams to work on this branch/mergerequest.

wilzbach:
I am taking the liberty as this has nothing to do with the codebase and I am giving this experiment a shot. If CodeCov turns out to be unstable, we can always disable this.

Issue #147, opened 2017-06-16
Very bad performance when parsing huge array literals.


BBasile:
The problem is illustrated with https://github.com/UplinkCoder/d_primes/blob/master/source/primes_4m.d, which contains a huge array initializer (16 Millions expressions !).

with this code:

```d
void main(string[] args)
{
    LexerConfig config;
    RollbackAllocator alloc;
    StringCache cache = StringCache(StringCache.defaultBucketCount);
    ubyte[] source = cast(ubyte[]) read("/home/basile/Dev/dproj/audited/repos/d_primes/source/primes_4m.d");

    auto toks = getTokensForParser(source, config, &cache);
    auto modl = parseModule(toks, "", &alloc);
}
```
I takes 2 minutes.

with this one:

```d
void main(string[] args)
{
    LexerConfig config;
    RollbackAllocator alloc;
    StringCache cache = StringCache(0xFFFF);
    ubyte[] source = cast(ubyte[]) read("/home/basile/Dev/dproj/audited/repos/d_primes/source/primes_4m.d");

    auto toks = getTokensForParser(source, config, &cache);
    auto modl = parseModule(toks, "", &alloc);
}
```

I takes 16 seconds.

- DCD and DScanner should be tweaked to initialize more smartly the bucket count (based on the file size maybe).
- In libdparse the performances of the StringCache should be checked.

It's probably also why this issue has been reported: https://github.com/dlang-community/DCD/issues/357

PR #148, opened 2017-06-16
Make the StringCache created in byToken more performant

BBasile:
This is related to #147. The default bucket count is anticipated using the count of byte to scan.

### DIFF
--- a/src/dparse/lexer.d
+++ b/src/dparse/lexer.d
@@ -1786,8 +1786,11 @@ private pure nothrow @safe:
  */
 public auto byToken(ubyte[] range)
 {
+    import std.math : nextPow2;
+    uint bc = cast(uint)((range.length > 2^^31UL) ? 2^^31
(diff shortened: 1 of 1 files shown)


BBasile:
trivial

src/dparse/lexer.d
```diff
@@ -1786,8 +1786,11 @@ private pure nothrow @safe:
  */
 public auto byToken(ubyte[] range)
 {
+    import std.math : nextPow2;
```
MartinNowak: Please be a bit careful with relying on newly introduced functions (2.071), in particular such trivial ones.

ZombineDev: Well, the oldest version of Phobos that we're testing is the on`ldc` ships with - i.e. 2.073.1. Even GDC master now supports this. I think we can assume that nothing older than that it's supported.

PR #149, opened 2017-06-19
Allow configureable WhitespaceBehavior in getTokensForParser

wilzbach:
Needed for https://github.com/dlang-community/D-Scanner/pull/448 and and https://github.com/dlang-community/D-Scanner/pull/450

I am not sure whether this is the best solution as it requires allocation by the user as `const(Token)[]` is 

```d
	tokens = getTokensForParser(code, config, &cache);
	auto tokensC = tokens; // <-- this can be used by the Dscanner checks
	if (config.whitespaceBehavior == WhitespaceBehavior.include)
		tokensC = tokens.filter!(t => t.type != WhitespaceBehavior.include).array;
...
	dparse.parser.parseModule(tokensC, fileName, p, report, ...)
...
```

The parser currently doesn't check for whitespace tokens and imho it would unnecessarily complicate its logic to add checks.

### DIFF
--- a/src/dparse/lexer.d
+++ b/src/dparse/lexer.d
@@ -384,9 +384,10 @@ public bool isLiteral(IdType type) pure nothrow @safe @nogc
 }
 
 /**
- * Returns: an array of tokens lexed from the given source code to the output range. All
- * whitespace tokens are skipped and comments are attached to the token nearest
- * to them.
+ * Returns: an array of tokens lexed from the given source code to the output range.
+ * Whitespace token are skipped by default, except `WhitespaceBehavior.include`
+ * has been set explicitly.
+ * comments are attached to the token nearest to them.
  */
 const(Token)[] getTokensForParser(ubyte[] sourceCode, LexerConfig config,
     StringCache* cache)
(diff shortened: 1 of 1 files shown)


wilzbach:
(rebased to see how CodeCov turns out to work)

src/dparse/lexer.d
```diff
@@ -425,6 +424,9 @@ const(Token)[] getTokensForParser(ubyte[] sourceCode, LexerConfig config,
     {
     case tok!"specialTokenSequence":
     case tok!"whitespace":
+        if (config.whitespaceBehavior == WhitespaceBehavior.include)
```
Hackerpilot: The (unchanged) doc comment specifically states that this won't happen.

wilzbach:
> The getTokensForParser function was given that name deliberately. The parser doesn't handle whitespace tokens so there should not be any whitespace in the value returned from this function.

Ok, but

1) What's the purpose of `config.whitespaceBehavior` then?
2) How do you propose to allow getting the whitespace tokens from libdparse then?(as e.g. needed for r dlang-community/D-Scanner#448 or dlang-community/D-Scanner#450)

Hackerpilot:
1. That config option is used inside of the lexer's popFront function.
2. It may be a good idea to attach the whitespace tokens to the non-whitespace tokens the way we do with comments.

WebFreak001:
there is a new PR (#387) which would replace this PR

WebFreak001:
closing because no longer needed

PR #150, opened 2017-06-22
Add support for C-style alias declarations

wilzbach:
In Phobos `std.typecons` and `std.traits` use this `alias` syntax:

```
 alias StorageClasses_opt BasicType FuncDeclarator ;
```

```
FuncDeclarator:
    BasicType2opt Identifier FuncDeclaratorSuffix
```

 As AFAICT there's no alternative syntax to this, replacing the few uses sadly doesn't work.
Hence, it would be very appreciated if Dscanner (and thus libdparse) could parse these declarations at least to the degree of not throwing a hard error ;-)
I gave this a shot and it looks like it's very hard to support.

References:
- https://dlang.org/spec/declaration.html#alias
- https://dlang.org/spec/function.html#FuncDeclarator

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -441,7 +441,8 @@ final class AliasDeclaration : ASTNode
 public:
     override void accept(ASTVisitor visitor) const
     {
-        mixin (visitIfNotNull!(storageClasses, type, identifierList, initializers));
+        mixin (visitIfNotNull!(storageClasses, type, identifierList, initializers,
+                               parameters, memberFunctionAttributes));
     }
     mixin OpEquals;
     /** */ StorageClass[] storageClasses;
@@ -449,6 +450,8 @@ public:
     /** */ IdentifierList identifierList;
     /** */ AliasInitializer[] initializers;
     /** */ string comment;
+    /** */ Parameters parameters;
+    /** */ MemberFunctionAttribute[] memberFunctionAttributes;
 }
 
 ///
--- a/README.md
+++ b/README.md
@@ -15,4 +15,3 @@ script. Running the tests on Windows is not currently supported.
 * [Static array initialization syntax](http://dlang.org/arrays.html#static-init-static). This syntax is ambiguous because it looks like an associative array literal.
 * [Class allocators](http://dlang.org/class.html#allocators). These are deprecated in D2.
 * [Class deallocators](http://dlang.org/class.html#deallocators). These are deprecated in D2.
-* C-style alias syntax. For example: ```alias int F1();```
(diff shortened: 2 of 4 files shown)


wilzbach:
See also: https://issues.dlang.org/show_bug.cgi?id=12299

src/dparse/ast.d
```diff
@@ -449,6 +449,8 @@ public:
     /** */ IdentifierList identifierList;
     /** */ AliasInitializer[] initializers;
     /** */ string comment;
+    /** */ Parameters parameters;
```
BBasile: You forget to visit the parameters

src/dparse/ast.d
```diff
@@ -449,6 +449,8 @@ public:
...
+    /** */ MemberFunctionAttribute[] memberFunctionAttributes;
```
BBasile: You forget to visit the memberFunctionAttributes

BBasile:
> As AFAICT there's no alternative syntax to this

Indeed. I've opend an issue about this https://issues.dlang.org/show_bug.cgi?id=16020 but i think that the only resolution possible is a personal initiative or a DIP. In the absence of the required language change supporting this in libdparse sounds like the right choice.

wilzbach:
> - Misses       4204    4210       +6

That stroke me as quite odd -> https://github.com/dlang-community/libdparse/pull/151

src/dparse/ast.d
wilzbach: My first PR to the parser. Thanks!! (added)

wilzbach:
> That stroke me as quite odd -> #151

Thanks a lot for the quick approval!
(Rebased to get a proper coverage report.)

wilzbach:
Okay @BBasile - how do we proceed here? Does this need approval from @Hackerpilot or can we move ahead?

BBasile:
> Does this need approval from @Hackerpilot

I've explained why i agree but initially not to parse them was a clearly a choice, so yes.

src/dparse/ast.d
Hackerpilot: Looks like this is fixed.

Hackerpilot:
I initially left support for this out as I thought
1. That the syntax would be deprecated / removed
2. That the new syntax was sufficient

wilzbach:
> I initially left support for this out as I thought
> That the syntax would be deprecated / removed
> That the new syntax was sufficient

Hehe, I think we all were/are still confused by this oversight...

So this means this PR is good to go? :)
I would love to get rid of the exclusion of `std.traits` and `std.typecons.d` [in the Phobos build](https://github.com/dlang/phobos/blob/6fe6dffdcd4262f0064bfd0dfc1c05225fd14ef8/posix.mak#L542) .

Hackerpilot:
Not sure if you noticed, but I clicked the "approve" button in Github.

wilzbach:
> Not sure if you noticed, but I clicked the "approve" button in Github.

I did, but I was just wondering whether there was a reason why you didn't hit merge afterwards ;-)
Thanks a lot!

PR #155, opened 2017-07-14
Fix deprecation: 'switch' skips declaration of variable

JohanEngelen:
Fixes this warning: `Deprecation: 'switch' skips declaration of variable std.d.parser.Parser.parsePrimaryExpression.ident`
I cannot test this well, but we hit this deprecation when compiling an older version of libdparse with LDC 1.3.0. See https://issues.dlang.org/show_bug.cgi?id=16254 and https://github.com/dlang/dmd/pull/5869 .

Thanks.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -4486,15 +4486,17 @@ class Parser
         case tok!"const":
         case tok!"inout":
         case tok!"shared":
-            advance();
-            expect(tok!"(");
-            mixin(parseNodeQ!(`node.type`, `Type`));
-            expect(tok!")");
-            expect(tok!".");
-            const ident = expect(tok!"identifier");
-            if (ident !is null)
-                node.primary = *ident;
-            break;
+            {
+                advance();
+                expect(tok!"(");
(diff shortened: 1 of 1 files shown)


src/dparse/parser.d
```diff
@@ -4486,15 +4486,17 @@ class Parser
...
-                node.primary = *ident;
-            break;
+            {
+                advance();
+                expect(tok!"(");
+                mixin(parseNodeQ!(`node.type`, `Type`));
+                expect(tok!")");
+                expect(tok!".");
+                const ident = expect(tok!"identifier");
+                if (ident !is null)
```
BBasile: Instead of creating a scope you can do

```d
if (auto ident = expect(tok!"identifier")
    node.primary = *ident;
```

JohanEngelen:
We are using our own branch, no quick merge needed. I prefer if you could fix it the way you'd like to.   (there are more cases where a variable is defined without a new scope, perhaps you want to add a scope there too regardless of the compiler complaining or not)

Issue #135, opened 2017-04-29
Missing `format` overrides


Geod24:
I have some generic code that generates visitors from all AST types, and as I was extending it I encountered this error:

```
./src/altered/visitor/ParseTree.d(122): Error: none of the overloads of 'format' are callable using argument types (const(AnonymousEnumDeclaration)), candidates are:
./submodules/libdparse/src/dparse/formatter.d(76):        dparse.formatter.Formatter!(Sink).Formatter.format(const(AddExpression) addExpression)
./submodules/libdparse/src/dparse/formatter.d(83):        dparse.formatter.Formatter!(Sink).Formatter.format(const(AliasDeclaration) aliasDeclaration, const(Attribute[]) attrs = null)
./submodules/libdparse/src/dparse/formatter.d(130):        dparse.formatter.Formatter!(Sink).Formatter.format(const(AliasInitializer) aliasInitializer)
./submodules/libdparse/src/dparse/formatter.d(147):        dparse.formatter.Formatter!(Sink).Formatter.format(const(AliasThisDeclaration) decl, const(Attribute[]) attrs = null)
./submodules/libdparse/src/dparse/formatter.d(161):        dparse.formatter.Formatter!(Sink).Formatter.format(const(AlignAttribute) alignAttribute)
./src/altered/visitor/ParseTree.d(122):        ... (189 more, -v to show) ...
```

Geod24:
Actually, it's:
- AnonymousEnumDeclaration
- AnonymousEnumMember
- AutoDeclarationPart

Basile-z:
Don't fix this one, rather make #346.

PR #159, opened 2017-07-23
fix #158 - handle identifier list containing an indexing expression

### LINKED ISSUE #158: Parsing type expressions involving array indexing is broken
T[1].X y;

BBasile:


BBasile:
The fix is wrong, `T[0].T[1].T` is well parsed but only the last indexer is stored. The problem is that the indexer cannot be parsed in `parseIdentifierOrTemplateInstance` because it is used everywhere and would have consequences  on the others constructs like the expressions.

In background the problem is that some kind of simplifications make the good fix harder than it looks.
https://dlang.org/spec/grammar.html#IdentifierList is not right in dparse.

BBasile:
I've made a nicer fix. As expected i had to create a new AST node type, so that each indexer of the chain is stored. As expected too all the tools break: D-Scanner, D-Symbol. The user tools that used Type2 will also break.

There's no other way to do that. I repeat myself but even if the bug was trivial, libdparse simplifies the AST so that things can be re-used (`IdentifierOrTemplateChain` here). One use case was forgotten and with the fix the IdentifierOrTemplateChain would not be reusable anymore.

Finally, if this get accepted i think that we should schedule this week, two hours during which the other tools could be fixed (because it requires each a time a review, a tag, etc).

BBasile:
This good now. Summary of the changes:

- `IdentifierList` is renamed to `DeclaratorIdentifierList`. It was used in alias declarations and in D specs, this is the name that's used: (https://dlang.org/spec/grammar.html#DeclaratorIdentifierList)
- `Type2` doesn't use an `IdentifierOrTemplateInstanceChain` but rather what's officialy an `IdentifierList` and that can include an indexer after each item of the list. This is the issue that's fixed in this PR.
- LinkageAttribute also uses this new `IdentifierList`, following the specs (https://dlang.org/spec/grammar.html#LinkageAttribute).

src/dparse/formatter.d
```diff
@@ -3867,3 +3903,35 @@ protected:
         "postblit"
     ];
 }
+
+version (unittest)
+void testFormatNode(Node)(string sourceCode)
```
BBasile: use this new function to fix #135 and also some problems in harbored-mod (which uses dparse formatter apparently...)

src/dparse/parser.d
```diff
@@ -3133,30 +3161,46 @@ class Parser
...
     /**
-     * Parses an IdentifierList
+     * Parses an IdentifierList.
      *
      * $(GRAMMAR $(RULEDEF identifierList):
-     *     $(LITERAL Identifier) ($(LITERAL ',') $(LITERAL Identifier))*
-     *     ;)
+     *      $(RULE identifierOrTemplateInstance)
+     *    | $(RULE identifierOrTemplateInstance) $(LITERAL '.') $(RULE identifierList)
+     *    | $(RULE identifierOrTemplateInstance) $(LITERAL '[') $(RULE assignExpression) $(LITERAL ']') $(LITERAL '.') $(RULE identifierList)
```
Hackerpilot: The language spec and I disagree on the topic of "IdentifierList" being only a list of identifiers.

Hackerpilot: That and the idea of "lists" being separated by commas.

BBasile: It's a chain, clearly.

Hackerpilot: I'm not objecting to your change. It's consistent with the spec. I'm just upset with the spec.

BBasile: There must be "chain" in the name. It's not a list.

BBasile: The only other use is with `extern(cpp)`.

src/dparse/formatter.d
```diff
@@ -3867,3 +3903,35 @@ protected:
...
+
+    class CatchInterestingStuff : ASTVisitor
+    {
+        alias visit = ASTVisitor.visit;
+        override void visit(const(Node) stuff)
+        {
+            stuff.accept(this);
+            format(&fmt, stuff);
+            assert(fmt.data.canFind(code), fmt.data);
+            writeln(fmt.data);
```
BBasile: gotta remove this writeln.

BBasile:
reverted. The changes are hard to handle.

PR #161, opened 2017-08-02
Fix #158 - handle indexer in type identifier chain

### LINKED ISSUE #158: Parsing type expressions involving array indexing is broken

BBasile:
This PR fix the bug, that is a type can be specified using an AliasSeq or a tuple, however it breaks the tools that rely on dparse.


BBasile:
By the way, dont merge this if you don't want to fix the tools. It's not an easy task.

BBasile:
he he that's the shit right ;)

stefan-koch-sociomantic:
It's fine to add a new Symbol; but please don't remove the existing one.

BBasile:
The problem is that it's wrongly named. I've opened an issue from the D side as well: https://issues.dlang.org/show_bug.cgi?id=17700

BBasile:
Since nobody seems motivated enough to handle this (because of the required changes in dsymbol i suppose) maybe in a first time the construct could be handled but ignored, meaning that `T[0].Y` would parse but will still be handled as `T.Y`.

BBasile:
@Hackerpilot:

I managed to update DSymbol for this change and was able to compile DCD. So i think that the PR can be merged. Two notes:

- as @stefan-koch-sociomantic noted this is a breaking change. There's no way not to break things.
- about DSymbol: in a first time the indexed components will be ignored but i hope to handle tuples and Alisaseq in a near future.

BBasile:
You can see that the conversion was actually very simple:

https://github.com/BBasile/dsymbol/commit/286ae579eaece00f9f7705119f9e05014ed0a9ae

BBasile:
> It's fine to add a new Symbol; but please don't remove the existing one.

Oh you didn't speak about the chain VS list thing. Actually yes i'll put deprecated aliases to simplify the transition.

Hackerpilot:
@stefan-koch-sociomantic How badly is this going to break things for you?

BBasile:
A trick to minimize the impact of the breaking changes would be to put a _0.8.0-alpha.1_ tag after this change (instead of _0.7.2-alpha.7_). DUB will continue using the 0.7.X releases, unless _~> 0.8.0-alpha.1_ is set explicitly in the description.

When used as a submodule the problem does not exists. People can just checkout the last SHA that's good for them.

Hackerpilot:
Yes. This is certainly the kind of change that requires a 0.7 to 0.8 version increment.

BBasile:
Warning, do not merge for now. I'm rewritting. What i named chain is an array of node.

BBasile:
I have to close that and reopen another PR. I had commit directly in the origin.

Issue #162, opened 2017-08-02
parse error for default parameter value of typesafe variadic


SSoulaimane:
the following should work:
```d
void fun(int i = 1, int[] rest = null ...)
{
}
```

PR #163, opened 2017-08-02
fix issue #162

SSoulaimane:
parse the following case:
```d
void fun(int i = 1, int[] rest = null ...)
{
}
```


BBasile:
Can you put the case that failed in a d file and ping to be sure nobody forgets this fix ? Thanks.

BBasile:
okay thank you for the fix.

BBasile:
do you need a tag for code.dlang.org ?

PR #164, opened 2017-08-17
test with min required compiler

MartinNowak:
Recently libdparse backwards compatibility with older than 2.071 phobos was broken for a simple nextPow2 usage.
https://github.com/dlang-community/libdparse/pull/148/files
Better be explicit and aware what the oldest supported D version is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ d:
   - dmd-nightly
   - dmd-beta
   - dmd
+  - dmd-2.071.0
   - ldc-beta
   - ldc
 sudo: false


skl131313:
I don't think it's reasonable to support such an old version of the compiler, and that's only the *minimum* version. Why aren't any of the other versions of the compiler being tested between 2.071 and the current version? I think 2.073.2 would be more reasonable, and if someone is using an older compiler then they can either update their compiler or issue a pull request to fix any compatibility issues themselves.

BBasile:
It guaranteed GDC compatibility. Maybe if Travis supports GDC then we should rather replace the old dmd by "gdc"

skl131313:
Seems the last release of GDC was still using 2.068 and the current repo seems to be updated to 2.075? Where does 2.071 come from?

BBasile:
from nowhere. it was about nextPow2 actually.

PR #166, opened 2017-08-20
fix #165 - An EponymousTemplateDeclaration is never parsed

### LINKED ISSUE #165: An EponymousTemplateDeclaration is never parsed
There seems to be a method/rule called `EponymousTemplateDeclaration`. The documentation provides a grammar specification, however it seems to be different from the official D definition of an Eponymous Template. It also appears that declarations like `enum foo(T) = bar!T;`, are not parsed as an `EponymousTemplateDeclaration`, but instead parsed as  `AutoDeclarationPart`. Is this a bug or/and when should I expect a EponymousTemplateDeclaration to be parsed?

BBasile:


BBasile:
this one is quite trivial, maybe @rikkimax or @skl131313 could have a look ? The problem is that there was a conflict between storage classes. `enum` alone is seen as `auto` so the special processing for `enum`, which comes later, was dead code.

BBasile:
Okay so verified in DCD and D-Scanner now. will merge.

PR #168, opened 2017-08-21
fix #156 - Parser bug with assoc-array literal

### LINKED ISSUE #156: Parser bug with assoc-array literal
The readme states that static array initialization syntax is unsupported, the following does not appear to be the same as the linked example, and when parsed, gives parse errors:

` string[uint[3]] aa1 = [[1,2,3]: "onetwothree"];  `

BBasile:
The issue is fixed by supporting static array initialization in most cases. Why ?

- associative array literals are less frequent since they require constant expressions.
- the index in static array initializers is usually a simple primary (char literal or int literal).

Based on these two facts we can take a descision when there's an ambiguity.
There are still cases of errors because a bit of semantic is actually required to be 100% exact.
In case of errors, the static array init is in an assoc array literal. I believe that these cases could be
handled in DSymbol.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,9 @@ Online documentation is available [here](https://dlang-community.github.io/libdp
 Tests are present in the test directory. To run them execute the run\_tests.sh
 script. Running the tests on Windows is not currently supported.
 
+# Differences with the official grammar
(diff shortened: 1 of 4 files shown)


BBasile:
I tested the other tools and good news: the changes neither break DCD (+DSymbol) nor D-Scanner. There could be consequences on user-code but probably not much.

Issue #157, opened 2017-07-18
Parse static foreach


BBasile:
It was merged in DMDFE yesterday !

BBasile:
@SSoulaimane [learn how to use GH issues](https://help.github.com/articles/closing-issues-using-keywords/)

PR #169, opened 2017-09-01
parse static foreach patch for issue #157

SSoulaimane:
parse static foreach #157 which is now in dmd official release since 2.076.0

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -2097,6 +2097,8 @@ class Parser
                 mixin (nullCheck!`node.conditionalDeclaration = parseConditionalDeclaration(strict)`);
             else if (peekIs(tok!"assert"))
                 mixin(parseNodeQ!(`node.staticAssertDeclaration`, `StaticAssertDeclaration`));
+            else if (peekIs(tok!"foreach") || peekIs(tok!"foreach_reverse"))
+                mixin(parseNodeQ!(`node.staticForeachDeclaration`, `StaticForeachDeclaration`));
             else
                 goto type;
             break;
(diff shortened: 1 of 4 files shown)


src/dparse/parser.d
```diff
@@ -4966,9 +5031,11 @@ class Parser
                 mixin(parseNodeQ!(`node.conditionalStatement`, `ConditionalStatement`));
             else if (peekIs(tok!"assert"))
                 mixin(parseNodeQ!(`node.staticAssertStatement`, `StaticAssertStatement`));
+            else if (peekIs(tok!"foreach") || peekIs(tok!"foreach_reverse"))
+                mixin(parseNodeQ!(`node.staticForeachStatement`, `StaticForeachStatement`));
             else
             {
-                error("'if' or 'assert' expected.");
+                error("'if' or 'assert' or 'foreach' or 'foreach_reverse' expected.");
```
Hackerpilot: ```error("'if', 'assert', 'foreach', or 'foreach_reverse' expected.");```

SSoulaimane: > I'd like to see a test case that checks that the new alias and enum attributes are parsed for the foreach statement.

Thank you @Hackerpilot for the review. I didn't know there were new type attributes.

src/dparse/parser.d
```diff
@@ -2864,29 +2864,32 @@ class Parser
...
-            node.isRef = true;
-            advance();
+            IdType typeConstructor;
+            if (currentIs(tok!"ref"))
+            {
+                node.isRef = true;
+                advance();
+            }
+            else if (currentIs(tok!"alias"))
+                advance(); // TODO: where should I put this
```
SSoulaimane: I'm not sure where the new attributes should be added. Should I make a new field `IdType[] attributes`? if so should it include ref? should it just be an alias to `typeConstructors` field? Should I add two more flags `bool isEnum` and `bool isAlias`? Or should it be a `IdType enumOrAlias` field?

Hackerpilot: That's a good question. Adding the extra isEnum and isAlias fields would mean that there aren't any backwards-incompatible changes to the AST classes, so I think that's the better thing to do even thought it might not result in the cleanest design.

wilzbach:
Tagged this as v0.7.2-alpha.1

PR #173, opened 2017-10-21
fix #170 - Function literal parsed as struct initialization

### LINKED ISSUE #170: Function literal parsed as struct initialization
The following:
```d
int function() func = {
    return 6;
};
```

parses as;

```xml
<module>
    <declaration>
        <variableDeclaration>
            <type pretty="int function()">
                <type2>int</type2>
                <function/>
                <parameters>
                </parameters>
            </type>
            <declarator line="1">
                <name>func</name><initializer>
                <nonVoidInitializer>
                    <structInitializer>
                        <structMemberInitializers>
                        </structMemberInitializers>
                    </structInitializer>
                </nonVoidInitializer>
            </initializer>
            </declarator>
        </variableDeclaration>
    </declaration>
</module>
```

Which does not seem to be correct.

BBasile:
The fix consists into checking if there's a ExpressionStatement inside the curly braces and now
it works exactly like for DMDFE, even corner cases:

```d
auto f = (){call();}; // worked already.
auto f = {call();}; // this was the bug, okay now as function literal.

// corner cases also fixed
auto f = {}; // struct init, even if semantically wrong
auto f = {;}; // func literal, even if empty statement
```

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -3489,3 +3489,38 @@ unittest // issue #156
     }
 }
 
+unittest // issue #170
+{
+    import dparse.lexer, dparse.parser, dparse.rollback_allocator;
+
+    final class Test170_F : ASTVisitor
+    {
+        static ubyte[] src = cast(ubyte[])q{void function() a = {call();};};
+        bool visited;
+        alias visit = ASTVisitor.visit;
+        override void visit(const FunctionLiteralExpression){visited = true;}
+    }
+
+    final class Test170_S : ASTVisitor
+    {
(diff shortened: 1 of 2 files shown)


BBasile:
Note that this only solves the issue when the type is explicit and non-aliased.

BBasile:
other possible fix: if a semicolon is found between the curly braces then it's a function literal otherwise it's a struct init.

BBasile:
Parsing now follow DMDFE, even for corner cases:

```d
auto f = {call();}; // this was the bug
// corner cases also fixed
auto f = {}; // struct init, even if semantically wrong
auto f = {;}; // func literal, even if empty statement

```

PR #177, opened 2017-11-05
fix #176 - infinite loop on align syntax error inside AsmStatement

### LINKED ISSUE #176: infinite loop on align syntax error inside AsmStatement
This crashes all the tools based on dparse, when this error is made:

```d
asm
{
    align 2);
}
```

BBasile:
Please once merged also push the `v0.7.2-alpha.2` tag. D-Scanner and DCD need to get the dependency upgraded.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -532,9 +532,20 @@ class Parser
             advance(); // align
             node.hasAlign = true;
             if (currentIsOneOf(tok!"intLiteral", tok!"identifier"))
+            {
                 node.identifierOrIntegerOrOpcode = advance();
+                if (!currentIs(tok!";"))
+                {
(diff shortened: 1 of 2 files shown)


src/dparse/parser.d
```diff
@@ -532,9 +532,20 @@ class Parser
             advance(); // align
             node.hasAlign = true;
             if (currentIsOneOf(tok!"intLiteral", tok!"identifier"))
+            {
                 node.identifierOrIntegerOrOpcode = advance();
+                if (!currentIs(tok!";"))
+                {
+                    error("';' expected.");
+                    advance();
+                    return null;
```
BBasile: The loop was `while (stack.put(<the result of this>){ ... }`. The temp node was always returned.

PR #180, opened 2017-11-20
Allow do as keyword for body statements

wilzbach:
I also tested this with the do-body PR at Phobos https://github.com/dlang/phobos/pull/5869


BBasile:
@wilzbach, to update dscanner you may need to push a tag here...

PR #182, opened 2017-12-12
make libdparse compatible with dmd v2.070.2

stefan-koch-sociomantic:
This pr is need for dfmt to be integrated with dmd version sociomantic is using.

### DIFF
--- a/src/dparse/lexer.d
+++ b/src/dparse/lexer.d
@@ -1780,13 +1780,20 @@ private pure nothrow @safe:
     bool haveSSE42;
 }
 
+/// copy from phobos b/c we need to build on older versions of dmd
+/// Returns : the next power of two from a given value
+private static size_t nextPow2(size_t value)
+{
+    import core.bitop : bsr;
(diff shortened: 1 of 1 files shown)


wilzbach:
Why do you need to support such old D versions? What's wrong with using a modern DMD to build dfmt?

wilzbach:
(that's why it's often helpful to use the PR description text)

leandro-lucarella-sociomantic:
I'm almost sure this wasn't really necessary, sorry for the misunderstanding. Please ping me to confirm we really need hacks like these, because normally we shouldn't.

wilzbach:
> I'm almost sure this wasn't really necessary, sorry for the misunderstanding. Please ping me to confirm we really need hacks like these, because normally we shouldn't.

Oh I asked Stefan on Slack and he told me there is no easy way for you guys to use a more modern compiler.
Well, if you don't need this, I think we should revert it...

@stefan-koch-sociomantic ?

PR #185, opened 2017-12-24
improve coverage and fix cases of unreachable errors

BBasile:
parser.d goes from 88% to 90%. A few problems were detected, see my own review comments for more details.


src/dparse/parser.d
```diff
@@ -3428,11 +3407,7 @@ class Parser
         }
 
         mixin(tokenCheck!")");
-        if (currentIs(tok!"}"))
```
BBasile: just above, the `mixin` returns so this branch was never executed

src/dparse/parser.d
```diff
@@ -3123,10 +3105,7 @@ class Parser
         if (isTemplate && currentIs(tok!"if"))
             mixin(parseNodeQ!(`node.constraint`, `Constraint`));
 
-        if (currentIs(tok!";"))
```
BBasile: `parseFunctionBody` returns if the first token is a semicolon. I removed here because `parseFunctionBody` may be called by hand at some time and for some reasons

src/dparse/parser.d
```diff
@@ -1423,9 +1408,6 @@ class Parser
         case tok!"immutable":
             node.first = advance();
             break;
-        default:
```
BBasile: `parseCastQualifier` is always called after a `isCastQualifier` check.  I removed the default case and added an assertion in case someone would one day call this sub parser alone.

src/dparse/parser.d
```diff
@@ -1213,9 +1198,8 @@ class Parser
     BodyStatement parseBodyStatement()
     {
         mixin(traceEnterAndExit!(__FUNCTION__));
+        assert(currentIsOneOf(tok!"body", tok!"do"));
         auto node = allocator.make!BodyStatement;
-        if (!currentIsOneOf(tok!"body", tok!"do"))
```
BBasile: Not covered because the sub parser is always called after a check.

src/dparse/parser.d
```diff
@@ -753,36 +753,21 @@ class Parser
...
-            case "qword":
-                break;
-            default:
-                error("ASM type node expected");
-                return null;
-            }
-            if (currentIs(tok!"identifier") && current().text == "ptr")
-                node.right = advance();
-            return node;
-        default:
```
BBasile: Not covered because the sub parser is always called after a check.

PR #188, opened 2018-01-24
Use std.conv.hexString instead of the deprecated language hex strings

wilzbach:
I'm going to auto-merge this as it currently breaks CircleCi on Phobos. I also added `disallowDeprecations` to prevent such problems in the future.


WebFreak001:
um I don't think it's a good idea to add `disallowDeprecations`. dmd will probably deprecate something in here eventually and then DCD and all the other programs depending on libdparse will break

wilzbach:
> um I don't think it's a good idea to add disallowDeprecations. dmd will probably deprecate something in here eventually and then DCD and all the other programs depending on libdparse will break

It's exactly the opposite. This _will_ prevent this from happening again. The Project Tester ensures it  - it runs libdparse, but fine I will separate this into another PR so you have more time to convince yourself that this is a good idea ;-)

wilzbach:
> I will separate this into another PR so you have more time to convince yourself that this is a good idea 

https://github.com/dlang-community/libdparse/pull/189

PR #189, opened 2018-01-24
Use -de for libdparse

wilzbach:
This will prevent annoyances like CircleCi (see e,g. [1]) on Phobos getting broken as the Project Tester uses `-de` and thus PRs at dmd can't break libdparse.

[1] https://github.com/dlang-community/libdparse/pull/188

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -4,4 +4,5 @@
 	"targetType": "library",
 	"description": "Library for lexing and parsing D source code",
 	"license": "BSL-1.0",
+	"buildRequirements": ["disallowDeprecations"],
 }
--- a/test/run_tests.sh
+++ b/test/run_tests.sh
@@ -18,7 +18,7 @@ echo "Running unit tests..."
 echo -e "${GREEN}DONE${NORMAL}"
 
 echo -en "Compiling tester... "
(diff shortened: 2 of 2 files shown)


MartinNowak:
To clarify the comment from https://github.com/dlang-community/libdparse/pull/188#issuecomment-360294934

>> um I don't think it's a good idea to add disallowDeprecations. dmd will probably deprecate something in here eventually and then DCD and all the other programs depending on libdparse will break
> 
> It's exactly the opposite. This will prevent this from happening again. The Project Tester ensures it - it runs libdparse, but fine I will separate this into another PR so you have more time to convince yourself that this is a good idea ;-)

We don't want to restrict anyone from adding a deprecation to dmd or phobos.
Deprecations are non-breaking changes that produce leeway in a coupled ecosystem, so when they are introduced in dmd or phobos, every maintainer can resolve the deprecations on their own schedule (within 6 month).
Thus the general idiom is to test your own code (but only your own code) with `-de -w`, so to notice unresolved deprecations/warnings, but use non-fatal deprecations and warnings for everybody elses code.

BBasile:
Personally i don't care, even if i use -de systematically too. 

Maybe another point of view before closing the PR ?

BBasile:
All right closing, lack of interest and conflicts.

PR #190, opened 2018-01-26
 fix #158 - handle identifier list containing an indexing expression

### LINKED ISSUE #158: Parsing type expressions involving array indexing is broken

BBasile:
See #161.


BBasile:
In the previous PR i said that i tried to split the new node in two: chain and nodes. This actually failed (for some reason killer.d started to compile again...) and was more complex than the current way to fix the issue so now the name of a type is made of `TypeIdentifierPart`s. `Part` also matches to the other nodes such as `AutoDeclarationPart` so i thinkg that this is a good name.

BBasile:
ping @Hackerpilot. Let's handle this now. Do not forget the tag please. DSymbol and DCD PRs will follow next.

BBasile:
You can merge since i cant anymore myself.

BBasile:
The tag here is expected too: https://github.com/dlang-community/dsymbol/pull/60/files#diff-406ece895a5b70a272cd76a2ef902836R6.

PR #194, opened 2018-02-01
fix #193 - Possible infinite loop when the type suffix is for an AA with a type starting with a TypeConstructor

### LINKED ISSUE #193: Possible infinite loop when the type suffix is for an AA with a type starting with a TypeConstructor
It impacts DCD and D-Scanner. Reduced std.traits trigering the bug: https://gist.github.com/BBasile/771797a053f168f769a0f1d854a28271

BBasile:
What happened when the bookmark was not restored is hard to reproduce so i test the AST.
Sorry but this will require again tags here, in dsymbol and PRs in D-Scanner, Dsymbol and DCD.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -3540,3 +3540,57 @@ unittest // issue #170
     assert(t170_s.visited);
 }
 
+unittest // issue #193
+{
+    import dparse.lexer, dparse.parser, dparse.rollback_allocator;
+
+    final class Test193 : ASTVisitor
+    {
+        static ubyte[] src = cast(ubyte[])q{const(Type1[immutable(Type2)]) qualAarray;};
+        size_t tc;
+        alias visit = ASTVisitor.visit;
+        override void visit(const TypeIdentifierPart tip)
+        {
+            tip.accept(this);
+            assert(tip.indexer is null);
+        }
+        override void visit(const Type type)
+        {
+            tc++;
+            if (tc == 1) //const(
+            {
+                assert(type.type2.typeConstructor == tok!"const");
+                assert(type.type2.typeIdentifierPart is null);
+            }
+            else if (tc == 2) //const(Type1
+            {
+                assert(type.typeConstructors.length == 0);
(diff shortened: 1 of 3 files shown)


src/dparse/ast.d
```diff
@@ -3540,3 +3540,57 @@ unittest // issue #170
...
+                assert(type.type2.typeIdentifierPart.identifierOrTemplateInstance
+                    .identifier.text == "Type1");
+                assert(type.typeSuffixes.length == 1);
+            }
+            else if (tc == 3) // immutable
+            {
+                assert(type.type2.typeConstructor == tok!"immutable");
+                assert(type.type2.typeIdentifierPart is null);
+            }
+            else if (tc == 4) // immutable(Type2
```
wilzbach: Typically switch is used for these things ;-)

src/dparse/parser.d
```diff
@@ -3304,8 +3304,13 @@ class Parser
             advance();
             if (!currentIs(tok!"]"))
             {
+                node.indexer = parseAssignExpression();
+                if (node.indexer is null)
+                {
+                    goToBookmark(b);
+                    return node;
```
wilzbach: This is never hit: 

image

BBasile: this is something else i have to put in fail files. Previously "uncoverage" was hidden by the mixin replaced by this fix.

BBasile: Actually there are plenty of granches made for errors that are not covered but working with files is boring... i gave a short eelier: see #185

BBasile:
Okay this is covered now. If you have between 30min to 1 hour free we could propagate this fix to the other products now.

wilzbach:
> If you have between 30min to 1 hour free 

You know it fails a bit like you left this org because you don't want to this work yourself ;-)
I tagged v0.8.0-alpha.2 - the bot should take care of the rest. I will have more time in the late evening.

BBasile:
Yeah actually last time the minor version of the deps has changed, not this time so this may work (except for the git submodules)

PR #197, opened 2018-02-04
Automatically build docs + grammar

wilzbach:
An experiment - this will require merging to try it out, but I'm fairly confident
that it will work (tested locally).

See also:
- https://github.com/dlang-community/DGrammar/issues/7
- https://gist.github.com/veelo/f7668510bad2e8c9212ab66104541fcc


wilzbach:
This is now live at http://libdparse.dlang.io and will be updated automatically in the future.

PR #209, opened 2018-02-07
Add quickstart example + open on run.dlang.io badge

wilzbach:
Just a start - someone can always propose a better example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,3 +20,50 @@ script. Running the tests on Windows is not currently supported.
 # Unsupported Syntax
 * [Class allocators](http://dlang.org/class.html#allocators). These are deprecated in D2.
 * [Class deallocators](http://dlang.org/class.html#deallocators). These are deprecated in D2.
+
+# Example
+
+```d
+/+dub.sdl:
+dependency "libdparse" version="~>0.7"
++/
+import dparse.ast;
+import std.stdio, std.range;
+
+class TestVisitor : ASTVisitor
+{
+    alias visit = ASTVisitor.visit;
+    int indentLevel;
+
+    override void visit(const FunctionDeclaration decl)
+    {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,3 +20,45 @@ script. Running the tests on Windows is not currently supported.
...
+import dparse.ast;
+import std.stdio;
+
+class TestVisitor : ASTVisitor
+{
+    alias visit = ASTVisitor.visit;
+
+    override void visit(const FunctionDeclaration decl)
+    {
+        decl.name.text.writeln;
```
BBasile: Put a nested function in the example and add `this.accept(decl);` here. It will show a bit the point of the visitor pattern (for example on run.dlang)

README.md
```diff
@@ -20,3 +20,45 @@ script. Running the tests on Windows is not currently supported.
...
+    }
+}
+
+void main()
+{
+    import dparse.lexer;
+    import dparse.parser : parseModule;
+    import dparse.rollback_allocator : RollbackAllocator;
+    import std.array : array;
```
BBasile: not needed, `parseModule` can take directly the result of `getTokensForParser`.

README.md
wilzbach: Good idea! I also added an `indentLevel` to make this even clearer.

README.md
```diff
@@ -20,3 +20,50 @@ script. Running the tests on Windows is not currently supported.
...
+    import dparse.lexer;
+    import dparse.parser : parseModule;
+    import dparse.rollback_allocator : RollbackAllocator;
+    import std.string : representation;
+
+    auto sourceCode = q{
+        void foo() @safe {
+            void bar();
+        }
+    }.dup;
```
wilzbach: Imho, ideally `.dup` isn't needed here, but that's a different issue - https://github.com/dlang-community/libdparse/issues/210

PR #212, opened 2018-02-15
allow lexer to directly take all types of strings and ubyte arrays, close #210

### LINKED ISSUE #210: Libdparse should work on const(ubyte)[] arrays
Add least `getTokensForParser` needs adjustment:

```
onlineapp.d(35,37): Error: function `dparse.lexer.getTokensForParser(ubyte[] sourceCode, LexerConfig config, StringCache* cache)` is not callable using argument types `(immutable(ubyte)[], LexerConfig, StringCache*)`
```

BBasile:


src/dparse/lexer.d
```diff
@@ -388,8 +389,8 @@ public bool isLiteral(IdType type) pure nothrow @safe @nogc
  * whitespace tokens are skipped and comments are attached to the token nearest
  * to them.
  */
-const(Token)[] getTokensForParser(ubyte[] sourceCode, LexerConfig config,
-    StringCache* cache)
+const(Token)[] getTokensForParser(R)(R sourceCode, LexerConfig config, StringCache* cache)
+if (is(Unqual!(ElementEncodingType!R) : ubyte) && isDynamicArray!R)
```
wilzbach: Why not relax `isDynamicArray` to `isRandomAccessRange`?

src/dparse/lexer.d
```diff
@@ -497,13 +498,14 @@ public struct DLexer
...
-    this(ubyte[] range, const LexerConfig config, StringCache* cache,
+    this(R)(R range, const LexerConfig config, StringCache* cache,
         bool haveSSE42 = sse42()) pure nothrow @safe
+    if (is(Unqual!(ElementEncodingType!R) : ubyte) && isDynamicArray!R)
     {
         this.haveSSE42 = haveSSE42;
         auto r = (range.length >= 3 && range[0] == 0xef && range[1] == 0xbb && range[2] == 0xbf)
             ? range[3 .. $] : range;
-        this.range = LexerRange(r);
+        this.range = LexerRange(cast(const(ubyte)[]) r);
```
wilzbach: You can use `representation` from `std.string` which does the same (though that's just a style nit)

src/dparse/lexer.d
BBasile: At the end LexerRange stores a `const(ubyte)[]` so it cannot be something else than a dynamic array.

BBasile: Actually this could always be tweaked but it's really not worth.

src/dparse/lexer.d
BBasile: Because `std.string` is not imported i won't change this.

README.md
```diff
@@ -49,16 +49,15 @@ void main()
...
         }
-    }.dup;
+    };
     LexerConfig config;
     auto cache = StringCache(StringCache.defaultBucketCount);
-    auto tokens = getTokensForParser(sourceCode.representation, config, &cache);
+    auto tokens = getTokensForParser(sourceCode, config, &cache);
 
     RollbackAllocator rba;
     auto m = parseModule(tokens, "test.d", &rba);
```
wilzbach: You would need to update the URL too, but I can do this in a quick follow-up too.
OTOH we would then need to release libdparse 0.8 because run.dlang.io only downloads the latest, stable version.

src/dparse/lexer.d
wilzbach: Fair enough.

PR #215, opened 2018-02-25
fix grammar for opaque struct & unions, add several missing FILE_FULL…

BBasile:
…_PATH


BBasile:
still not in the auto generated grammar.

BBasile:
@wilzbach, this is still not in the auto generated grammar. I wonder if it has ever worked since i remember that you had to update by hand several times.

PR #217, opened 2018-04-01
Allows incomplete switch statement, better for DCD

BBasile:
Will fix https://github.com/dlang-community/DCD/issues/453 automatically


BBasile:
i'll push a patch tag. As discussed we should stop using too much the alpha/beta as optional semver component.

PR #219, opened 2018-04-22
doc, add more missing special KW, sort them and fix a few tabs

BBasile:
Dparse followed https://dlang.org/spec/grammar.html#SpecialKeyword but the list is incomplete. For example `__DATE__` is also accepted as default template parameter.


BBasile:
I think that the bot fails because of codecov, which is wrong here. gonna merge by hand.

PR #221, opened 2018-04-24
for DCD, return valid node on incomplete func call '(}'

BBasile:
This will fix https://github.com/dlang-community/DCD/issues/405.


BBasile:
I start believing that there should be a version identifier for the code dedicated to make DCD better.
What do you think @Hackerpilot ?

Hackerpilot:
I think that any time we accept invalid code we should still generate an error message. Doing this for DCD would be fine because it ignores errors from the parser anyways.

PR #222, opened 2018-04-24
update to stdx-allocator 2.77.1

bogdan-szabo-sociomantic:


BBasile:
BTW no need to update the other JSON ~> 2.77.0 is compat with ~>2.77.1

PR #228, opened 2018-05-05
 fix #227 - Gagged error in static if but should be emitted

### LINKED ISSUE #227: Gagged error in static if but should be emitted
The following code parses w/o error:

```d
static if (true)
{
    template A()
    {
        enum a = 1
        enum b = 0;
    }
}
```

remove the static if and the error is back

BBasile:


BBasile:
@Hackerpilot: why did you put a bookmark in first place ? If it was important i can store and restore the index so that errors are still emitted.

Hackerpilot:
Bookmarks are usually used when the parser cannot determine ahead of time if it is looking at a declaration or a statement.

Hackerpilot:
The old behavior may be incorrect since it seems that at this point in the code the parser has already determined that it is a declaration.

BBasile:
Actually i think the only case where there can be an ambiguity is for `DeclarationOrStatement`, which already handles error gagging before calling `parseDeclaration` (and indirectly `parseConditionalDeclaration`). So the bookmark in `parseConditionalDeclaration` was not necessary i think.

PR #230, opened 2018-05-07
Always update the git submodules

wilzbach:
Project Tester is still failing with:

```
+ [ ! -d stdx-allocator ]
+ cd test
+ ./run_tests.sh
find: ‘../stdx-allocator/source’: No such file or directory
Compiling unit tests...../src/dparse/parser.d(9): Error: module `mallocator` is in file 'stdx/allocator/mallocator.d' which cannot be read
```

So I assume this is due to the directory being existent, but being empty.

See also: https://github.com/dlang/ci/pull/205

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
           - libevent-dev
 script:
   # note: for dlang project tester, which doesn't clone automatically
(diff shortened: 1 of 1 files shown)


wilzbach:
Tagged as v0.8.3

wilzbach:
libdparse is now re-enabled on the project tester (:

PR #231, opened 2018-05-11
[trivial] Fix range violation error

s-ludwig:


stefan-koch-sociomantic:
please add a testcase, thanks.

PR #238, opened 2018-06-03
Workaround the DMD bug causing corruption while parsing types

BBasile:


BBasile:
Let's say that if https://issues.dlang.org/show_bug.cgi?id=18941 is not closed on September the 1st we'll merge the workaround.

PR #242, opened 2018-06-14
fix #241 - `body` should be supported as identifier

### LINKED ISSUE #241: `body` should be supported as identifier

BBasile:


test/tester.d
```diff
@@ -295,7 +295,7 @@ int main(string[] args)
         config.fileName = arg;
         const(Token)[] tokens = getTokensForParser(fileBytes, config, &cache);
         RollbackAllocator rba;
-        parseModule(tokens, arg, &rba, &messageFunction);
+        parseModule(tokens, arg, &rba, toDelegate(&messageFunction));
```
Geod24: That looks unrelated, isn't it ?

BBasile: indeed.

BBasile: removed

PR #244, opened 2018-07-05
Support for DIP1009 (new contracts syntax), close #218

### LINKED ISSUE #218: Support for DIP1009 (new contracts syntax)
See https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1009.md#grammar.
No an emergency since it's not yet implemented so not used at all.

BBasile:
The changes are implemented without huge AST modification. `InContractExpression` and `OutContractExpression` are lowered to `InStatement` and `OuStatement`. For now the grammar in the DDOC comments still needs to be updated and once done it wont be nice, because of the lowering.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -3012,7 +3012,7 @@ class Parser
      *
      * $(GRAMMAR $(RULEDEF functionBody):
      *       $(RULE blockStatement)
-     *     | ($(RULE inStatement) | $(RULE outStatement) | $(RULE outStatement) $(RULE inStatement) | $(RULE inStatement) $(RULE outStatement))? $(RULE bodyStatement)?
+     *     | ($(RULE inStatement) | $(RULE outStatement))* $(RULE bodyStatement)?
      *     ;)
      */
     FunctionBody parseFunctionBody()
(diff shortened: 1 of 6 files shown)


BBasile:
invariant are still to do

BBasile:
invariants done too.

BBasile:
So for now while everything is parsed correctly, rendered grammar is not complete and would hardly be.
It's still possible to re-write using the official rules (even if DMD does not either). So let's discuss before merging. With summer, nice weather, etc i confess that i'm not super motivated... this could change  in a few weeks or maybe if someone wants to do it properly, then do it.

src/dparse/parser.d
```diff
@@ -3024,26 +3024,94 @@ class Parser
...
+                else return null;
             }
             else if (currentIs(tok!"out"))
             {
-                mixin(parseNodeQ!(`node.outStatement`, `OutStatement`));
-                if (currentIs(tok!"in"))
-                    mixin(parseNodeQ!(`node.inStatement`, `InStatement`));
+                if (next.type == tok!"{" || (index < tokens.length - 3  &&
+                    tokens[index + 2].type == tok!("identifier") &&
+                    tokens[index + 3].type == tok!(")")))
```
Hackerpilot: I really should add a `peekIs` overload that takes an offset to avoid verbose code like this.

src/dparse/parser.d
```diff
@@ -3024,26 +3024,94 @@ class Parser
...
+        if (currentIs(tok!"{") && !requireDo)
+        {
+            mixin(parseNodeQ!(`node.blockStatement`, `BlockStatement`));
+        }
+        else if (currentIs(tok!";"))
+        {
+            advance();
+            return node;
+        }
+        else if (currentIsOneOf(tok!"do") || current.text == "body")
```
Hackerpilot: Should be `currentIs` if only one argument is needed.

src/dparse/parser.d
Hackerpilot: Wrote https://github.com/dlang-community/libdparse/issues/246 as a reminder for later. This isn't something that needs to be changed for this pull request to be merged.

BBasile:
I've managed to add an almost up to date grammar in DDOC. It's correct but it doesn't expresses the case where **do** is not required (only some {in|out}ContractExpressions before body) because it's not possible if new rules are not 1:1 layered on the DIP.

PR #250, opened 2018-07-22
fix #31 - Follow issue 13943 (super and this as type)

### LINKED ISSUE #31: Follow issue 13943
https://issues.dlang.org/show_bug.cgi?id=13943

BBasile:


BBasile:
this would be the first deprecation message in the parser. Maybe the issue can be closed withuot merging this. Important thing is that the issue is handled now. We'll remove the deprecated stuff when they will be removed of DMD

PR #257, opened 2018-08-08
Add Meson build definition

eclipseo:
Based on @ximion work on https://github.com/dlang-community/containers/commit/fc97b589e485d6e80e6f17c828abaa6c9fb27f35

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,23 @@ matrix:
     addons:
       apt:
         packages:
+          - pkg-config
           - libevent-dev
+
+install:
+  - pyenv global system 3.6
+  - pip3 install 'meson==0.47'
+  - mkdir .ntmp
+  - curl -L https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-linux.zip -o .ntmp/ninja-linux.zip
+  - unzip .ntmp/ninja-linux.zip -d .ntmp
+
+before_script:
+  export PATH=$PATH:$PWD/.ntmp
+
 script:
   # note: for dlang project tester, which doesn't clone automatically
+  - meson build && ninja -j8 -C build
+  - ninja -j8 -C build test -v
   - git submodule update --init --recursive
   #
   - cd test && ./run_tests.sh && cd ..
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,77 @@
+ project('dparse', 'd',
+    meson_version: '>=0.44',
+    license: 'BSL-1.0',
+    version: '0.9.3'
+)
+
+project_soversion = '0'
+
+pkgc = import('pkgconfig')
+allocator_dep = dependency('stdx-allocator', version: '>= 2.77', fallback: ['stdx-allocator', 'allocator_dep'])
+
+#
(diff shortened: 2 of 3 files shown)


wilzbach:
I think you should add it to `.travis.yml` too, see e.g. https://github.com/dlang-community/containers/blob/master/.travis.yml

eclipseo:
It failed because of a bug of stdx-allocator: https://github.com/dlang-community/stdx-allocator/issues/2

wilzbach:
Huh? That's just a deprecation warning. The real issue is that the current config passes the wrong flags to LDC:

```
Aborting from rt/sections_elf_shared.d(489) Only one D shared object allowed for static runtime. Link with shared runtime via LDC switch '-link-defaultlib-shared'.
```

.travis.yml
```diff
@@ -12,9 +12,23 @@ matrix:
     addons:
       apt:
         packages:
+          - pkg-config
           - libevent-dev
+
+install:
+  - pyenv global system 3.6
+  - pip3 install 'meson>=0.45'
```
wilzbach: Based on previous experiences with meson and its instability, I recommend locking the version.

eclipseo:
I'm confused on how to solve this error.

wilzbach:
> I'm confused on how to solve this error.

CC @ximion who is the D Meson expert.

eclipseo:
Finally it works, it needed a newer meson and `meson.get_compiler('d').get_id() == 'llvm'` and not `'ldc'`

It should be good now.

ximion:
That you have to pass `-link-defaultlib-shared` is very strange... Which Linux distribution do you use?

eclipseo:
>That you have to pass -link-defaultlib-shared is very strange... Which Linux distribution do you use?

I don't have for my distro (Fedora), only because of Travis.

meson.build
```diff
@@ -0,0 +1,77 @@
...
+    dependencies: [allocator_dep]
+)
+
+#
+# Tests
+#
+if meson.get_compiler('d').get_id() == 'llvm'
+  extra_args = ['-main', '-link-defaultlib-shared']
+else
+  extra_args = ['-main']
```
ximion: This will fail with GDC, to be compatible with all D compilers, it's usually better to create an explicit main() function somewhere in the code.
Or do this:
```d
version(unittest) void main () {}
```

BBasile:
Excepted that 0.9.3 was released today so finally it's worth rebasing and change the version number.

eclipseo:
@BBasile I rebased it and updated for 0.9.3.

BBasile:
Thanks but sorry, there will be 0.9.4 today i think. A major error has been reveled when D-Scanner started to use 0.9.3 and nothing should be built with this version (parser can fall into an infinite loop).

BBasile:
I've updated the version to 0.9.4 and merging this time. Where will be the D stuff ? official repos ? Fusion ?

eclipseo:
@BBasile official, though I'm not the packager, only the package reviewer. @bioinfornatics will be your packager.

PR #259, opened 2018-08-08
Deploy via netlify

wilzbach:
As GitHub pages seem to be permanently broken.


wilzbach:
All links will be the same. The main difference is just that netlify works really well and even comes with a few gimmicks (like SSL or PR documentation preview).

BBasile:
thanks.

PR #262, opened 2018-08-20
Add support for UDAs on function arguments, close #261

### LINKED ISSUE #261: Add support for UDAs on function arguments
https://forum.dlang.org/thread/pl79jr$2nio$1@digitalmars.com?page=1

BBasile:


src/dparse/ast.d
```diff
@@ -2429,6 +2430,18 @@ public:
...
 }
 
+final class ParameterAttribute : ASTNode
+{
+    override void accept(ASTVisitor visitor) const
+    {
+        mixin (visitIfNotNull!(atAttribute));
+    }
+
+public:
```
Hackerpilot: Isn't this redundant? I think that the default for class members is `public`. Putting it here just implies that `accept` isn't `public`.

BBasile: Removed. I will remove them from all the ASTNodes later

PR #268, opened 2018-08-29
fix #267 - Infinite loop caused by the `in` keyword

### LINKED ISSUE #267: Infinite loop caused by the `in` keyword
```d
void main()
{
    in
}
```

Never returns.

BBasile:
Tested to fix the problem in dastworx.


BBasile:
bot's on strike

PR #273, opened 2018-09-05
provide token location to expression-based contracts

LaurentTreguier:
This is for [dfmt#375](https://github.com/dlang-community/dfmt/issues/375); this PR allows dfmt to properly recognize expression-based contracts as contracts since offsets of tokens are used for this.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -3079,10 +3079,14 @@ class Parser
                 }
                 else if (next.type == tok!"(")
                 {
+                    const size_t inTokenLocation = current.index;
                     advance();
                     advance();
                     if (InStatement s = parseContractExpression!InStatement())
+                    {
+                        s.inTokenLocation = inTokenLocation;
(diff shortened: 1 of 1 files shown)


src/dparse/parser.d
```diff
@@ -3079,10 +3079,14 @@ class Parser
                 }
                 else if (next.type == tok!"(")
                 {
+                    size_t inTokenLocation = current.index;
```
BBasile: `const size_t` please

src/dparse/parser.d
```diff
@@ -3104,6 +3108,7 @@ class Parser
                 }
                 else if (next.type == tok!"(")
                 {
+                    size_t outTokenLocation = current.index;
```
BBasile: here too

BBasile:
I suppose that you will need the 0.9.7 tag ?

LaurentTreguier:
For dfmt's dub build, yes.

PR #283, opened 2018-10-12
fix #282 - Function parameter attributes are ignored

### LINKED ISSUE #282: Function parameter attributes are ignored
When using an `ASTVisitor`, the `ParameterAttribute` object in `void visit(const ParameterAttribute parameterAttribute)` will always have a null `atAttribute` property even if there is an attribute attached to the parameter.

An easy way to observe this is using D-Scanner:
```sh
echo "@(42) void foo(@(42) int a);" | ./bin/dscanner --ast | xmllint --format -
```
produces
```xml
<?xml version="1.0"?>
<module>
  <declaration>
    <attribute>
      <atAttribute>
        <argumentList>
          <unaryExpression>
            <primaryExpression>
              <intLiteral>42</intLiteral>
            </primaryExpression>
          </unaryExpression>
        </argumentList>
      </atAttribute>
    </attribute>
    <functionDeclaration line="1">
      <name>foo</name>
      <type pretty="void">
        <type2>void</type2>
      </type>
      <parameters>
        <parameter>
          <name>a</name>
          <type pretty="int">
            <type2>int</type2>
          </type>
          <identifier>a</identifier>
        </parameter>
      </parameters>
    </functionDeclaration>
  </declaration>
</module>
```

The `<attribute>` tag and its contents are present for the function attribute, but missing for the parameter.

BBasile:


Geod24:
No test ?

BBasile:
Coverage report says that the LOC is executed, good enough isn't it ?

BBasile:
https://codecov.io/gh/dlang-community/libdparse/pull/283/diff

PR #286, opened 2018-10-14
fix #147, fix #285, better performance with huge array initializers

### LINKED ISSUE #147: Very bad performance when parsing huge array literals.

### LINKED ISSUE #285: AutoDeclaration storage classes are not handled by the formatter

BBasile:
An new parser property allows to take only one array element per dimension of an `ArrayInitializer`, which has for effect to save the amount of memory used by the AST. This property will be set depending on the product, e.g DSymbol yes (just need one to infer auto arrays), D-Scanner AST printer no (need accurate AST), etc.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -3433,12 +3433,55 @@ unittest // there used to be a small regression when adding the ParserConfig
     import dparse.lexer, dparse.parser, dparse.rollback_allocator;
 
     auto src = q{module m;};
+    RollbackAllocator ra;
+    LexerConfig cf = LexerConfig("", StringBehavior.source);
+    StringCache ca = StringCache(16);
+    Module m1 = parseModule(getTokensForParser(src, cf, &ca), "", &ra , null);
+    Module m2 = parseModule(getTokensForParser(src, cf, &ca), "", &ra);
+}
+
+unittest
+{
+    import dparse.lexer, dparse.parser, dparse.rollback_allocator;
+    import dparse.formatter, std.array, std.string;
+
+    final class TestArrayInitVisitor : ASTVisitor
+    {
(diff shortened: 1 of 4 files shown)


BBasile:
Need a few change after #288.

BBasile:
Updated so that the new `ParserConfig` is used.

BBasile:
too much changes would be required to parse correctly without creating the nodes of the array init.

PR #288, opened 2018-10-14
Add a ParserConfig struct, close #287

### LINKED ISSUE #287: Add a ParserConfig struct
`parseModule` had too much parameter now.

BBasile:
Why such a radical change ?

- the struct allows to pass a config created from a `StructInitializer`, which is the only versatile way that's not based on N overloads. Most of the time the current code is compatible, i.e in existing code just surround parseModule parameters with `ParserConfig()`, **easy maintenance**.
-  both a function and a delegate message function are allowed (no more deprecation), which will allow to remove all the `toDelegate` which were added last year.
- The change allows **nice UFCS style expressions** : `ParserConfig(...).parseModule();`
- Nothing is deprecated because of point 1 and also because this change would be part of 0.10.x, i.e **no breaking changes** if the rule of thumb that's to use the `~>` version operator in DUB json is respected.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -17,50 +17,86 @@ import std.string : format;
 // Caution: generates 180 megabytes of logging for std.datetime
 //version = dparse_verbose;
 
+
+/**
+ * Prototype for a custom parser message function or delegate.
+ * Params:
+ *      fileName = The source file name.
+ *      line = The line in the source, 0 based.
+ *      column = The column in the source, 0 based.
+ *      message = Datailed message.
+ *      isError = Indicates if the message is a warning (`false`) or a if it's an error (`true`).
+ */
+alias MessageFunction = void function(string fileName , size_t line, size_t column, string message, bool isError);
+
+/// ditto
+alias MessageDelegate = void delegate(string, size_t, size_t, string, bool);
+
+/**
+ * Parser configuration struct
+ */
+struct ParserConfig
+{
+    /// The tokens parsed by dparse.lexer.
+    const(Token)[] tokens;
+    /// The name of the file being parsed
+    string fileName;
+    /// A pointer to a rollback allocator.
+    RollbackAllocator* allocator;
+    /// An optional function used to handle warnings and errors.
+    MessageFunction messageFunction;
+    /// An optional delegate used to handle warnings and errors.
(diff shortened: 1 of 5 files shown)


src/dparse/parser.d
```diff
@@ -17,49 +17,61 @@ import std.string : format;
...
-    if (errorCount !is null)
-        *errorCount = parser.errorCount;
+    with (parserConfig)
+    {
+        parser.fileName = fileName;
+        parser.tokens = tokens;
+        parser.messageFunction = messageFunction;
+        parser.messageDelegate = messageDelegate;
+        parser.allocator = allocator;
+    }
```
Hackerpilot: This `with` block makes it look like the Parser should have a reference to the ParserConfig struct, since the configuration struct just seems to have duplicate copies of the same fields.

BBasile: Yes but access to the tokens would have an additional indirection.

Hackerpilot: Could that be solved with an `alias this`?

BBasile: No the issue is about the supplemental dereference that would happen each time `current()` is called for example. But for this one now i think that it can be copied from the ParserConfig to the Parser, exceptionally. The other config member won't because they're less used.

BBasile:
The 3rd commit adds a reference to a config in the parser but IMO it's not as a nice.

BBasile:
Reverted

BBasile:
ping @Hackerpilot and maybe others for decision.
ping @wilzbach for netlify.

wilzbach:
The error was valid:


```
src/dparse/parser.d(30,1): Warning: Ddoc: function declaration has no parameter 'fileName'
2:29:42 PM: src/dparse/parser.d(30,1): Warning: Ddoc: function declaration has no parameter 'line'
2:29:42 PM: src/dparse/parser.d(30,1): Warning: Ddoc: function declaration has no parameter 'column'
2:29:42 PM: src/dparse/parser.d(30,1): Warning: Ddoc: function declaration has no parameter 'message'
2:29:42 PM: src/dparse/parser.d(30,1): Warning: Ddoc: function declaration has no parameter 'isError'
```

BBasile:
This will be fixed in #297 because another PR is required anyway for d-symbol

PR #290, opened 2018-10-15
fix #289 - Inconsistencies of IfStatement conditions

### LINKED ISSUE #289: Inconsistencies of IfStatement conditions
I was testing the following cases using `ASTVisitor` and looking at the AST produced for the condition between parens:
```d
if (auto i = 1) {} // 1
if (const i = 1) {} // 2
if (int i = 1) {} // 3
if (const int i = 1) {} // 4
if (const(int) i = 1) {} // 5
```
The first thing is: unlike the others, condition `2` is parsed as an `AssignExpression`, meaning that the resulting `IfStatement.identifier` will not contain the identifier, and `IfStatement.expression` contains everything instead.

The second thing is: `3` and `4` will produce exactly the same result. The `const` qualifier will not be taken into account, since there is currently no property holding type constructors in `IfStatement`.
I don't know what the best way to resolve this would be; maybe simply adding `IdType[] typeConstructors` to `IfStatement`, like with `ForeachType`, would be sufficient ?

Both of these can be verified quickly using `dscanner --ast`.

BBasile:
In addition it's been verified that [case 2](https://github.com/dlang-community/libdparse/issues/289#issuecomment-429669330) now works in DCD.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -3560,58 +3560,75 @@ class Parser
     IfStatement parseIfStatement()
     {
         mixin(traceEnterAndExit!(__FUNCTION__));
-        auto node = allocator.make!IfStatement;
+        IfStatement node = allocator.make!IfStatement;
         node.line = current().line;
         node.column = current().column;
         mixin(tokenCheck!"if");
         if (moreTokens)
             node.startIndex = current().index;
         mixin(tokenCheck!"(");
+        const b = setBookmark();
 
-        if (currentIs(tok!"auto"))
+        // auto identifier = exp
+        if (moreTokens && currentIs(tok!"auto") && peekIs(tok!"identifier"))
         {
+            abandonBookmark(b);
             advance();
-            const i = expect(tok!"identifier");
-            if (i !is null)
-                node.identifier = *i;
-            expect(tok!"=");
+            node.identifier = advance();
+            mixin(tokenCheck!"=");
             mixin(parseNodeQ!(`node.expression`, `Expression`));
         }
-        else
-        {
-            // consume for TypeCtors = identifier
-            if (isTypeCtor(current.type))
-            {
-                while (isTypeCtor(current.type))
-                {
-                    const prev = current.type;
-                    advance();
-                    if (current.type == prev)
-                        warn("redundant type constructor");
-                }
-                // goes back for TypeCtor(Type) = identifier
-                if (currentIs(tok!"("))
-                    index--;
-            }
 
-            immutable b = setBookmark();
-            immutable c = allocator.setCheckpoint();
-            auto type = parseType();
-            if (type is null || !currentIs(tok!"identifier")
-                || !peekIs(tok!"="))
+        // const shared ...
+        if (!node.expression && moreTokens && isTypeCtor(current.type))
+        {
+            StackBuffer idTypes;
+            while (moreTokens)
             {
-                allocator.rollback(c);
-                goToBookmark(b);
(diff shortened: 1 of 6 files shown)


src/dparse/parser.d
```diff
@@ -3560,60 +3560,82 @@ class Parser
...
         mixin(tokenCheck!"if");
         if (moreTokens)
             node.startIndex = current().index;
         mixin(tokenCheck!"(");
+        Bookmark b = setBookmark();
 
-        if (currentIs(tok!"auto"))
+        // most commn case: rel expression and such
+        if (Expression e = parseExpression())
+            if (moreTokens && currentIs(tok!")"))
```
BBasile: The need for this reveals a bug:

    if (int i = a < 100) b(); 

An expression is found before reaching the right paren.

BBasile:
This one really needs review. Apparently parsing `IfStatement` was turned very hard because of a bug (see 3rd commit).

test/pass_files/templates.d
```diff
@@ -4,11 +4,11 @@ template a(b, c) if (d) {}
 template a(alias int a : 10) {}
 template a(alias int a : int) {}
 template a(alias int a : 10) {}
-template a(alias int a : 10 = int) {}
+//template a(alias int a : 10 = int) {}
 template a(alias int a : int = 10) {}
 template a(alias int a : int) {}
 template a(alias int a = 10) {}
-template a(alias int a = int) {}
+//template a(alias int a = int) {}
```
BBasile: this one was valid, so the problem is not fixed

BBasile: it passes when uncommented. Problem was well fixed actually.

src/dparse/parser.d
```diff
@@ -4896,6 +4918,7 @@ class Parser
             }
             else if (currentIs(tok!"("))
                 mixin(parseNodeQ!(`node.arguments`, `Arguments`));
+            else goto default;
```
BBasile: BasicType could be accepted as expression whatever followed.

LaurentTreguier:
`if (const i = 1) {}` is now parsed correctly, however I'm getting weird results with `typeCtors`...
I tried:
```d
override void visit(const IfStatement ifStatement)
{
    writef("%s ", ifStatement.typeCtors.length);
    foreach (ctor; ifStatement.typeCtors)
        if (ctor != tok!"")
            writef("%s ", str(ctor));
    write('\n');
}
```
With the input:
```d
void foo()
{
    if (const var = something)
        doIt();

    if (const int var = something)
        doIt();
}
```
And got the output:
```
80 .. !<= ( const
80 /= !<= floatLiteral const
```
I don't see how this could happen in the code though, so maybe it's something on my end.

BBasile:
Yeah, the output makes no sense.

BBasile:
writef("%d ", ifStatement.typeCtors.length);

BBasile:
yeah that's the issue.

BBasile:
@LaurentTreguier  Can you test again please ?

BBasile:
@LaurentTreguier  Can you pull latest changes and test again please ?

LaurentTreguier:
It looks all good for me now, the length is 1 and the only type constructor is const :+1:

BBasile:
@LaurentTreguier, when a PR is ok for you you can approve formally : 

formal_approval

BBasile:
Not good finally. There are plenty of errors when parsing parser.d for example. For now i don't know if it's because of error gagging or if parsing is really broken.

BBasile:
`if (T t = T.init)`, parser thinks T is an expression.

PR #292, opened 2018-10-19
Fix issue 291: do not require do{} for in{} followed by in()

FeepingCreature:
The only thing that matters is the nature of the last pre/postcondition. If it's a {}, you need do{}; if it's a (), you don't.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -3084,6 +3084,7 @@ class Parser
                 }
                 else if (next.type == tok!"(")
                 {
+                    requireDo = false;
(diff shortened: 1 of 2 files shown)


BBasile:
You need to check if [SimpleParser](https://github.com/dlang-community/dsymbol/blob/master/src/dsymbol/conversion/package.d#L147) in symbol.d is affected too. Those two must always be updated at the same time.

FeepingCreature:
It looks like dsymbol is just completely broken and will probably develop all-new unrelated error cases around out(foo; bar) { }.

PR #300, opened 2018-11-01
Prevent running CI twice when PR branch is in origin

BBasile:


BBasile:
Should work better after https://github.com/rejectedsoftware/ddox/pull/216

PR #302, opened 2018-11-02
Avoid more cases of possibly segaulting `current.type`

BBasile:


BBasile:
Coverage is red because this is mostly a protection measure in case of. There was no attempt to get it green at all.

Issue #303, opened 2018-11-03
Implement DIP1009 correctly


BBasile:
Meaning without lowering. See official grammar.

BBasile:
Maybe @LaurentTreguier you would agree to do it ? I remember that you didn't like current implementation when you have worked on DFMT last month.
Otherwise anyone else you want to just connect to GH and put yourself as Assignee. Would be great if done before the 9 of November.

LaurentTreguier:
I'll have a look and see what can be done.
If I understand correctly (sometimes, I don't), you're talking about parsing expression-based contracts as their own constructs and not transforming them into `InStatement`s and `OutStatement`s as it's currently done ?

BBasile:
yes, correct DIP 1009 would be some thing like that.

PR #304, opened 2018-11-03
 Implement DIP1009 correctly (#303)

LaurentTreguier:
This PR adds the following changes from the API perspective:
- new `InContractExpression` and `OutContractExpression` classes to represent expression-based contracts
- new `inContractExpressions` and `outContractExpressions` properties for `FunctionBody` to store them
- new `assertion` and `message` properties for `Invariant` in the case of an expression-based  invariant

These are obviously breaking changes since they change the way `FunctionBody` and `Invariant` store expression-based contracts.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -242,6 +242,7 @@ abstract class ASTVisitor
     /** */ void visit(const ImportDeclaration importDeclaration) { importDeclaration.accept(this); }
     /** */ void visit(const ImportExpression importExpression) { importExpression.accept(this); }
     /** */ void visit(const IndexExpression indexExpression) { indexExpression.accept(this); }
+    /** */ void visit(const InContractExpression inContractExpression) { inContractExpression.accept(this); }
     /** */ void visit(const InExpression inExpression) { inExpression.accept(this); }
     /** */ void visit(const InStatement inStatement) { inStatement.accept(this); }
     /** */ void visit(const Initialize initialize) { initialize.accept(this); }
@@ -268,6 +269,7 @@ abstract class ASTVisitor
     /** */ void visit(const Operands operands) { operands.accept(this); }
     /** */ void visit(const OrExpression orExpression) { orExpression.accept(this); }
     /** */ void visit(const OrOrExpression orOrExpression) { orOrExpression.accept(this); }
+    /** */ void visit(const OutContractExpression outContractExpression) { outContractExpression.accept(this); }
     /** */ void visit(const OutStatement outStatement) { outStatement.accept(this); }
     /** */ void visit(const ParameterAttribute parameterAttribute) { parameterAttribute.accept(this); }
     /** */ void visit(const Parameter parameter) { parameter.accept(this); }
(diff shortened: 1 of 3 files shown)


LaurentTreguier:
I had based this on the way `AssertExpression` is handled, and didn't realize that the DIP changed its grammar as well.
I'll have a better look at the grammar and stick to it

LaurentTreguier:
I've tried to follow the changes of DIP1009 more closely.
I used the name `AssertArguments` instead of `ContractArguments` since it's the name used in the grammar on [dlang.org](https://dlang.org/spec/grammar.html#AssertArguments).

src/dparse/ast.d
```diff
@@ -1909,6 +1925,18 @@ final class IndexExpression : ExpressionNode
...
 }
 
+///
+final class InContractExpression : ASTNode
+{
+    override void accept(ASTVisitor visitor) const
+    {
+        mixin (visitIfNotNull!(assertArguments));
+    }
+    /** */ size_t inTokenLocation;
```
BBasile: I don't think this is required. IIRC you wanted this for DFMT but with the official grammar FunctionContract gives the information if it's a expression style contract (its InOutStatement is null) or a statement style contract (its InOutContractExpression) is null.

src/dparse/ast.d
```diff
@@ -2288,6 +2352,19 @@ final class OrOrExpression : ExpressionNode
...
 }
 
+///
+final class OutContractExpression : ASTNode
+{
+    override void accept(ASTVisitor visitor) const
+    {
+        mixin (visitIfNotNull!(parameter, assertArguments));
+    }
+    /** */ size_t outTokenLocation;
```
BBasile: same as comment related to InContractExpression

src/dparse/ast.d
LaurentTreguier: `inTokenLocation` and `outTokenLocation` were already used for detecting classic contract statements; the problem with them was simply that they weren't filled for expression-based contracts, so DFMT didn't consider them as contracts.
But `inTokenLocation` and `outTokenLocation` are used to distinguish contracts from other usages of `in` and `out`, and not to distinguish expression-based ones from statement ones, so I think it's probably better to leave them there

BBasile: ok then. since my comment i just remembered too that DFMT is based on the tokens and only very partially the AST.

src/dparse/parser.d
```diff
@@ -5201,6 +5227,48 @@ class Parser
...
+        {
+            if (auto c = parseFunctionContract())
+            {
+                requireDo = c.inOutStatement !is null;
+                contracts.put(c);
+            }
+        }
+        ownArray(node.functionContracts, contracts);
+
+        if (current.type == tok!"do"
```
BBasile: prefer `currentIs(tok!"do")`

src/dparse/parser.d
```diff
@@ -5201,6 +5227,48 @@ class Parser
...
+                || (current.type == tok!"identifier" && current.text == "body"))
```
BBasile: `if ( moreTokens && .. ` to prevent any RangeError.

src/dparse/parser.d
```diff
@@ -684,7 +684,7 @@ class Parser
             {
                 trace("Found register");
                 mixin (nullCheck!`(node.register = parseRegister())`);
-                if (current.type == tok!":")
+                if (currentIs(tok!":"))
```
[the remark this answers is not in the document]
LaurentTreguier: Oh ok.

PR #317, opened 2018-11-06
fix #316 - Regression, function literal parsed as struct initializer

### LINKED ISSUE #316: Regression: function literal parsed as struct initializer
As I was starting to update DFMT to libdparse 0.10.x, I stumbled upon a regression in libdparse: with the code `auto fun = {int i;};`, the braces seem to be considered as a struct initializer instead of a function literal, causing wrong formatting.
Git bisect tells me that this started with e6c83b2174ed41259be3caa8bbbd593f87ff87f0

BBasile:
There must be a problem because the fix is so much simpler than the previous mess.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -3441,9 +3441,33 @@ unittest // issue #170
         override void visit(const FunctionLiteralExpression){visited = true;}
     }
 
+    final class Test170_F1 : ASTVisitor
+    {
+        static string src = q{  const a = {int i;};  };
+        bool visited;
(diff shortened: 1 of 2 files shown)


LaurentTreguier:
For now the only remaining problem I see is with empty braces, they seem to be still parsed as function literals instead of as an empty struct initializer

BBasile:
Yes i have noted this case but it's not solvable without semantic i think. (D not context free ?)

LaurentTreguier:
I don't see any way to solve this either. This specific case can probably be handled by DFMT directly.

PR #322, opened 2018-11-20
fix #321 - __vector should be accepted without type

### LINKED ISSUE #321: __vector should be accepted without type
D grammar is incomplete for this but actually `if(is(T == __vector))` is valid.

BBasile:
Please review quickly, this is blocks a phobos PR.


BBasile:
ping @dlang-community/all, it's green.

PR #324, opened 2018-11-21
fix #323 - Add support for __traits as BasicType

### LINKED ISSUE #323: Add support for __traits as BasicType (D issue 7804)

BBasile:


src/dparse/parser.d
```diff
@@ -7374,6 +7379,16 @@ protected: final:
         if (!moreTokens()) return false;
         switch (current.type)
         {
+        case tok!("__traits"):
```
BBasile: This new block causes the bad coverage report because `isDeclaration` is surprisingly never used by the parser.
However i've updated it because it can be used by 3rd parts applications.

PR #328, opened 2018-12-05
Fix trailing semicolon of invariant seen as an empty statement

LaurentTreguier:
As the title says, in [expression-based invariants](https://dlang.org/spec/grammar.html#ClassInvariant), a trailing semicolon is needed, but apparently it was forgotten


BBasile:
CI looks broken due to Meson but didn't want to deal with that. Please one of the python / Meson person handle this.

wilzbach:
Just remove meson until @ximion has time to fix it?

PR #330, opened 2018-12-06
Try to fix CI

BBasile:


LaurentTreguier:
With `pyenv global system` it should be able to pick up the system-installed python (from https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-global)

BBasile:
Yeah but that didn't work. I don't remember in which commit i tried this. :laughing:

LaurentTreguier:
Hmm... Since it's complaining about pip3 not being installed, wouldn't it be fixed by adding `python3-pip` to the apt packages installed ?

BBasile:
pyenv doesn't recognize the version of python 3 that's setup by default (3.4.3). This must be a regression at TravisCI or pyenv. The CI passed when i setup manually another  python version.

BBasile:
I'll try because installing python is slow.

BBasile:
Better wait they fix their regression because this is a silly workaround :clown_face:

PR #331, opened 2018-12-06
Make all ast nodes contain the parsed tokens

### LINKED ISSUE #191: Add precise source locations (start and end) to each ASTNode
Many tools depend on having the source range of syntax nodes accessible. "Peek definition" popups in IDEs need them to avoid guessing. Refactoring operations need them to avoid destroying the code layout.

While there are no current tools profiting from this that I know of, adding this feature would open up new possibilities.

WebFreak001:
fix #191

All nodes now contain a `const(Token)[] tokens` which is a slice of the input tokens which contains all the tokens which were parsed for that AST node.

The first commit simply adds the tokens array to each type

The second commit replaces old startIndex/startLocation and endIndex/endLocation with getter properties dynamically getting from the tokens array. It is not fully backwards compatible as before the user could set the value and now only getting works. However they should all return the same values as in previous versions.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -418,7 +418,14 @@ template generateOpEquals(T)
         ~ "\treturn true;\n}\nreturn false;";
 }
 
-abstract class ExpressionNode : ASTNode
+abstract class BaseNode : ASTNode
+{
+    /** List of tokens consumed by this AST node */ const(Token)[] tokens;
+
+    abstract void accept(ASTVisitor visitor) const;
+}
+
+abstract class ExpressionNode : BaseNode
 {
     override void accept(ASTVisitor visitor) const
     {
@@ -447,7 +454,7 @@ final class AddExpression : ExpressionNode
 }
 
 ///
-final class AliasDeclaration : ASTNode
+final class AliasDeclaration : BaseNode
 {
     override void accept(ASTVisitor visitor) const
     {
@@ -465,7 +472,7 @@ final class AliasDeclaration : ASTNode
 }
 
 ///
-final class AliasInitializer : ASTNode
+final class AliasInitializer : BaseNode
 {
     override void accept(ASTVisitor visitor) const
     {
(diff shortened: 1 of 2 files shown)


BBasile:
I'm no convinced anymore about the idea of adding virtual methods. The point was to avoid duplicating fields because some nodes contain the info in a token but now i think that two fields would be better.

WebFreak001:
so should I revert the second commit? Everything will still contain all the tokens as array because I need the exact ranges of everything in a part of my tool

src/dparse/ast.d
```diff
@@ -418,7 +419,14 @@ template generateOpEquals(T)
         ~ "\treturn true;\n}\nreturn false;";
 }
 
-abstract class ExpressionNode : ASTNode
+abstract class BaseNode : ASTNode
```
Geod24: Why wasn't `ASTNode` changed instead of adding an extra layer ?

WebFreak001: ASTNode was an interface, which can't have fields. It would have broken code extending interfaces from ASTNode

WebFreak001:
I undid the property methods because dsymbol seems to be writing to startLocation and endLocation of BlockStatement and because that could be a separate PR.

I put my patches on https://gist.github.com/WebFreak001/12931809526a73abe5a245564d65df77 (also added some missing `const`s)

BBasile:
rebase on master please

WebFreak001:
@BBasile rebased & tested

BBasile:
It's necessary to check the memory impact. Your use case in d-code, which is something like a "introduce function a here menu item"  doesn't convince me so far, although it's true that to have the position in each node was an unrealted request. in x86_64 there will be 16 bytes in more for each node. There can be many nodes and i won't be surprised to see the mem usage jumping to 100 Mo in more.

WebFreak001:
I tested locally with 32 GB total memory, it grew from a memory usage of 112 MB to 148 MB.

Doesn't seem to matter if it indexes any symbols or not because "hello world", a vibe-d application and serve-d with basically all of dlang-community as dependencies have the same memory usage both before and after change...

So there seems to be a constant increase of memory usage by DCD of roughly 30% on my machine.

BBasile:
All right then

PR #341, opened 2019-01-16
fix #338 - A class Inheriting from an AliasSeq item is seen as an error

### LINKED ISSUE #338: A class Inheriting from an AliasSeq item is seen as an error
Hopefully this must be in class parsing, not type parsing

```d
import std.meta;
class Base {}
alias Seq = AliasSeq!(Base);
class C : Seq[0] {}     
```

> Expected `{` instead of `[`

BBasile:
`BaseClass` contained a `Type2` so the indexer, since not followed by a `.` was not parsed. Only `Type` can have an indexer followed by nothing. Also get rid of the `BaseClass` node since it was a useless indirection.

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -181,7 +181,6 @@ abstract class ASTVisitor
     /** */ void visit(const AutoDeclarationPart autoDeclarationPart) { autoDeclarationPart.accept(this); }
     /** */ void visit(const BlockStatement blockStatement) { blockStatement.accept(this); }
     /** */ void visit(const BreakStatement breakStatement) { breakStatement.accept(this); }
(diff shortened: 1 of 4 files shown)


Geod24:
Won't that break code though ?

BBasile:
Yes, i confirm the breakage. In dsymbol there will be changes to do. But dsymbol 0.6.z is not used for now so for this this doesn't matter. Next time DCD submodules will be updated this is for a new DCD version too (0.11.0).

BBasile:
Obviously dparse would see its minor version number incremented too.

BBasile:
official D grammar was wrong too, see https://issues.dlang.org/show_bug.cgi?id=19592. So on hold until the specs are fixed.

BBasile:
Will rather be fixed following the change of the dlang.org PR.

PR #345, opened 2019-01-27
fix a range violation while formating ConditionalDeclaration

Basile-z:
And fix a formatting error of `static if`


Basile-z:
This for https://github.com/dlang-community/harbored-mod/issues/100.

Issue #346, opened 2019-01-27
The formatter needs a global rewrite


Basile-z:
There are plenty of bugs.

Basile-z:
I advice any taker to get inspired by [this AST formatter](https://github.com/Basile-z/styx/blob/master/src/styx/ast_formatter.d)

- properly tested
- 100% coverage
- AST visitor usage prevent to forget items.

Although STYX is simpler and easier to format.

PR #357, opened 2019-04-07
fix #356 - Add support for `ref` in function literals

### LINKED ISSUE #356: Add support for `ref` in function literals
This will come (probably) as part of DMDFE 2.086, see https://github.com/dlang/dlang.org/pull/2618.

Basile-z:


src/dparse/parser.d
```diff
@@ -5282,6 +5294,14 @@ class Parser
             else
                 mixin(parseNodeQ!(`node.arrayLiteral`, `ArrayLiteral`));
             break;
+        case tok!"ref":
+            if (peekIs(tok!"("))
+            {
+                import std.stdio;
```
wilzbach: Remove

Basile-z:
I forgot to update the formatter

Basile-z:
This is good now but unfortunately this PR shows again that #346 is required. For now I just applied a few changes to fix the SEGFAULTs and small errors encountered to support the change to func literals.

PR #358, opened 2019-05-01
add support for the new function type alias syntax

Basile-z:
Will be added by https://github.com/dlang/dmd/pull/9667.


src/dparse/parser.d
```diff
@@ -5516,7 +5526,7 @@ class Parser
         mixin(tokenCheck!"shared");
         mixin(tokenCheck!"static");
         mixin(tokenCheck!"~");
-        return parseStaticCtorDtorCommon(node, startIndex);
```
Basile-z: error slipped in... this would revert #354

PR #359, opened 2019-05-11
prevent invalid scope when parsing bodies containing erroneous decls

Basile-z:
Required to fix ...  https://github.com/dlang-community/DCD/issues/600

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -2371,13 +2371,26 @@ class Parser
             if (!declarationsAndStatements.put(parseDeclarationOrStatement()))
             {
                 allocator.rollback(c);
+
+                // detect the pattern ".}" for DCD. This is what happens when
+                // located at the end of a well delimited body/scope and requesting
+                // completion. This is also a case where it's sure sure that
+                // there's no ambiguity, even if it happens during a lookup:
+                // it's not a decl, it's not a statement, it's an error.
+                if (currentIs(tok!"}") && index > 0 && previous == tok!".")
+                    break;
+
                 if (suppressMessages > 0)
                     return null;
+
                 // better for DCD, if the end of the block is reached then
                 // go back, allowing the following declarations to be in
                 // the right scope, instead of the block we were in.
                 if (index > 0 && previous == tok!"}")
+                {
                     index -= 1;
+                    break;
+                }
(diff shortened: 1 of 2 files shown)


Basile-z:
this breaks the arbitrary lookups required by the language.

WebFreak001:
if `parseDeclarationsAndStatements` can't return null anymore, I think the following classes/members would never be null anymore:
- `CaseStatement.declarationsAndStatement`
- `CaseRangeStatement.declarationsAndStatement`
- `DefaultStatement.declarationsAndStatement`

`BlockStatement.declarationsAndStatement` before was null with an error in the content, now it is really only null if it's empty (`}` following the `{`)

Because of this PR I also found the following potential null-reference-errors which would be fixed by this:

formatter.d:616 (`void format(const CaseRangeStatement caseRangeStatement)`) and formatter.d:638 (`void format(const CaseStatement caseStatement)`) both call `formatCaseDecls` without null check of `declarationsAndStatements` and the method itself just accesses members.

formatter.d:1036 (`void format(const DefaultStatement defaultStatement)`) directly accesses the member without null check

nvm stopping checking this out if it's deleted then

Basile-z:
Yeah the formatter bugs are not a surprise, see #346.

Basile-z:
@WebFreak001, I have maybe a working fix coming.

Basile-z:
Okay it seems to work and doesn't break dparse test suite. This time the approach is more pragmatic. The case that breaks DCD, i.e a syntax error, is exactly detected.

src/dparse/parser.d
```diff
@@ -2371,13 +2371,29 @@ class Parser
...
                 allocator.rollback(c);
+
+                // detect the pattern ".}" for DCD. This is what happens when
+                // located at the end of a well delimited body/scope and requesting
+                // completion. This is also a case where it's sure sure that
+                // there's no ambiguity, even if it happens during a lookup:
+                // it's not a decl, it's not a statement, it's an error.
+                if (currentIs(tok!"}") && index > 0 && previous == tok!".")
+                {
+                    advance();
```
Basile-z: must not advance

Basile-z:
DCD test suite work too now.

PR #369, opened 2019-08-16
lexer: remove unused newlineMask function

kubo39:


kubo39:
Hmm, this function seems to be never used...? https://github.com/dlang-community/libdparse/commit/238916741e253f7a4a5a86781afdc16367c98b37

Hackerpilot:
I'm sure I wrote that function for a reason, but it appears that you're right.

Hackerpilot (approved):

PR #374, opened 2020-01-04
Fix #372 - Error when parsing DMDs backend/ptrtab.d

### LINKED ISSUE #372: Error reported when parsing a DMD source file (backend/ptrntab.d)
I noticed this after seeing D-Scanner parse `src/dmd/backend/ptrntab.d` in DMD's source tree.
Starting from 4c0e52133f8d7ba5de2099be673c821fac04cd7a, this file suddenly becomes full of errors when scanning it.
The first reported error is [at line 4794](https://github.com/dlang/dmd/blob/ed9025c1bb01ae003a7be6f15d6e2e2354d0b9d9/src/dmd/backend/ptrntab.d#L4794), with supposedly a missing parenthesis.

Tested on Linux x86_64, compiling using DMD 2.089.0 and the distro's LDC 1.15.0, with DMD's 2.089.0 source.

### LINKED ISSUE #373: UDA supported?
In the documentation [https://libdparse.dlang.io/dparse/ast.html](https://libdparse.dlang.io/dparse/ast.html) I cannot find any information whether and how UDA are supported.

I would look to scan source code files and find e.g. classes which have a specific UDA assigned. Is this possible with libdparse?

MoonlightSentinel:
Credit goes to Basile B. for supplying this patch.


stefan-koch-sociomantic:
Please put the tests into dedicated files.
they are a little big for unit-tests

MoonlightSentinel:
@stefan-koch-sociomantic Okay, i guess they belong into `test/pass_files`?

MoonlightSentinel:
Don't know why `stdx-allocator / test_allocator` fails on travis but it seems unrelated

Hackerpilot:
I was looking through the proposed changes and couldn't figure out the relation between them and the file that the parser is rejecting. I did a bit more testing and discovered that the file parses without errors if there are 101 entries in the array. You can reproduce the error if there are 102 entries. There is something very strange going on here.

MoonlightSentinel:
> I was looking through the proposed changes and couldn't figure out the relation between them and the new test file. 

It adressess the changes introduced in cf12c2d64133a1412f9551e67f462956c56aa5ee which is the first faulty commit.

> I did a bit more testing and discovered that the new file parses without errors if there are 101 entries in the array. You can reproduce the error if there are 102 entries. There is something very strange going on here.

I've noticed this strange behaviour as well. My initial guess was some sort of overflow but i don't know this code base well enough to track it down. 

Feel free to propose a better solution,so that DMD can upgrade to a more recent D-Scanner version.

MoonlightSentinel:
Ping

Hackerpilot:
I think there's something wrong with the `parseNonVoidInitializer` function. I'll get to this when I have time.

Hackerpilot:
I was right about the number of struct initializers being the key to all of this. Bumping up the `MAX_ERRORS` constant fixes the problem. I have a better fix on the way.

Hackerpilot:
See https://github.com/dlang-community/libdparse/pull/377

PR #375, opened 2020-01-06
Fix a range violation in unDecorateComment

Hackerpilot:
lineSplitter and D's grammar disagree about which characters are considered line breaks, so it's possible for a comment that should be a single line to be split by lineSplitter, leading to the array slice failing. This only happens in strange cases where there are form feeds or vertical tabs in the comment.


src/dparse/lexer.d
```diff
@@ -1851,36 +1851,43 @@ in
...
+				? LineType.normal
+				: LineType.strange;
+			if (prevLineType == LineType.strange)
+			{
+				outputRange.put(line);
+				prevLineType = currentLineType;
+				break;
+			}
+			else
+				prevLineType = currentLineType;
```
WebFreak001: the previous code uses spaces but this code mixed in tabs now

src/dparse/lexer.d
```diff
@@ -1851,36 +1851,43 @@ in
...
-        j = comment.length;
-
-        foreach (line; lineSplitter(comment))
-        {
+        foreach (line; lineSplitter!(KeepTerminator.yes)(comment))
+        {
+			immutable LineType currentLineType = line.endsWith("\n")
+					|| line.endsWith("\r")
+					|| line.endsWith("\u2029")
+					|| line.endsWith("\u2030")
```
WebFreak001: you used paragraph separater and permille sign here, should be `\u2028` and `\u2029`

src/dparse/lexer.d
```diff
@@ -1851,36 +1851,43 @@ in
...
+				? LineType.normal
+				: LineType.strange;
+			if (prevLineType == LineType.strange)
+			{
+				outputRange.put(line);
+				prevLineType = currentLineType;
+				break;
```
WebFreak001: this should probably be `continue`?

src/dparse/lexer.d
```diff
@@ -1851,36 +1851,43 @@ in
...
-                outputRange.put('\n');
-            else
-                hasOutput = true;
-            if (stripped.length >= leadingChars.length && stripped.startsWith(leadingChars))
-                outputRange.put(stripped[leadingChars.length .. $]);
+            if (l.length >= leadingChars.length && l.startsWith(leadingChars))
+                outputRange.put(l[leadingChars.length .. $]);
             else
-                outputRange.put(stripped);
+                outputRange.put(l);
```
WebFreak001: for the last 4 lines: `outputRange.put(l.chompPrefix(leadingChars));`

src/dparse/lexer.d
```diff
@@ -1851,36 +1851,40 @@ in
...
+                : LineType.strange;
+            if (prevLineType == LineType.strange)
+            {
+                outputRange.put(line);
+                prevLineType = currentLineType;
+                continue;
+            }
+            else
+                prevLineType = currentLineType;
             auto l = line[3 .. $];
```
WebFreak001: I think this should much rather be
```suggestion
            auto l = line.chompPrefix("///");
```
and the continue from the "strange line type" if should actually be removed.

When there is a strange line ending, the next line will probably be what will be most corrupted, missing the starting `///` and with this maybe even range erroring. This also means it's not really needed to have separated logic and would massively simplify this code.

A `chompPrefix` (startsWith + slice) is probably more efficient than 4 `endsWith` checks anyway

PR #376, opened 2020-01-06
Fix some more range errors on unexpected EOF

Hackerpilot:


src/dparse/parser.d
```diff
@@ -521,6 +521,11 @@ class Parser
         mixin(traceEnterAndExit!(__FUNCTION__));
         auto startIndex = index;
         AsmBrExp node = allocator.make!AsmBrExp();
+        if (!moreTokens)
```
WebFreak001: this should be done before the allocator call so we don't potentially leak memory

PR #378, opened 2020-01-26
Fix trace(string) built with version=dparse_verbose

MoonlightSentinel:
suppressMessages was changed in 95c31d5a444a3c55d59a5f9a8b4378d48bdf23a1.

Should probably change the tag `v0.13.2` as well


MoonlightSentinel:
CC @Hackerpilot

PR #380, opened 2020-02-03
Add: Members to AST about existence of tokens

Sobaya007:


src/dparse/ast.d
```diff
@@ -2640,6 +2645,7 @@ final class SpecifiedFunctionBody : BaseNode
     }
     /** */ FunctionContract[] functionContracts;
     /** */ BlockStatement blockStatement;
+    /** */ bool hasDo;
```
WebFreak001: unused/unassigned in this PR

src/dparse/ast.d
```diff
@@ -894,6 +894,7 @@ final class AtAttribute : BaseNode
     /** */ ArgumentList argumentList;
     /** */ TemplateInstance templateInstance;
     /** */ Token identifier;
+    /** */ bool useParen;
```
WebFreak001: this could probably easily be made into a property function checking `this.tokens` to save some memory (and alignment bytes)

Sobaya007:
As you say, most braces are easily checked.
But I think it is hard to check the existence of "do" in `SpecifiedFunctionBody` and braces in `ConditionalDeclaration`.
The reason I added other changes is just to keep consistency.

I did not concern about saving memory, but can it be a problem to add just a few members?

WebFreak001:
in general we have been avoiding extra members the more common a node is. These won't hurt all that much but still for the easy ones they can much easier be implemented as property functions.

Though now I can think of a use case: modifying the members to affect formatting/serialization of AST to code. If you add checks for these members in the format function I am fine with these members existing

src/dparse/parser.d
```diff
@@ -1770,6 +1772,7 @@ class Parser
         if (currentIs(tok!":") || currentIs(tok!"{"))
         {
             immutable bool brace = advance() == tok!"{";
+            node.useBraceForTrueDeclarations = brace;
```
WebFreak001: for consistency this should be able to describe if current is `:`

Maybe add `enum StatementBlockStyle : ubyte { single, colon, block }` with some small ddoc explaining the items

src/dparse/ast.d
```diff
@@ -1622,7 +1625,7 @@ final class Foreach(bool declOnly) : BaseNode
     /** */ ForeachType foreachType;
     /** */ Expression low;
     /** */ Expression high;
-    /** */ bool useBrace;
+    /** */ StatementBlockStyle style;
```
WebFreak001: this should be moved into the `static if (declOnly)` block 2 lines below because in the other block there is already DeclarationOrStatement which can either only be a single thing or a block statement. (where you know it uses braces)

This is only set in the `static if (declOnly)` code block too, which means this would always be "single" otherwise

src/dparse/ast.d
```diff
@@ -75,6 +75,9 @@ shared static this()
     typeMap[typeid(XorExpression)] = 48;
 }
 
+/// Statement block style
```
WebFreak001: just repeating the name for something like this isn't really good documentation. I would suggest:

```suggestion
/// Describes which syntax was used in a list of declarations in the containing AST node
```

src/dparse/ast.d
```diff
@@ -75,6 +75,9 @@ shared static this()
...
+enum StatementBlockStyle : ubyte { single, colon, block }
```
WebFreak001: StatementBlockStyle wasn't a good name idea by me because there is already BlockStatement and other statements are single, so this should rather be `DeclarationListStyle` and also document its members because this is how it is used.
```suggestion
enum DeclarationListStyle : ubyte
{
	/// A declaration directly after the containing AST node making it the only child
	single,
	/// A colon (`:`) was used in the containing AST node meaning all following declarations are part here.
	colon,
	/// The declarations have been specified in a block denoted by starting `{` and ending `}` tokens.
	block
}
```

Sobaya007:
What should I do about conditional declaration which has ":"?
Should I parse it and add change about it?

WebFreak001:
hm no, make it a separate PR when needed (also needs tests then)

PR #381, opened 2020-02-09
Change RollbackAllocator.Node.contains to exclusive end

WebFreak001:
This fixes an issue in rollback(point) that the assert(contains(point)) would not error out but then it would crash eventually because first becomes null while recursively checking the contains.

Before it would keep deallocating even when the memory point we have is right outside the allocated memory region.

This shouldn't introduce any side effects as contains is only used in the rollback_allocator module and now it would simply assert out instead of segfaulting out.

(got libdparse to segfault in the rollback call in parseFunctionBody and this changes it to an assert fail but I'm still searching for the reason why it is invalid in the first place)

### DIFF
--- a/src/dparse/rollback_allocator.d
+++ b/src/dparse/rollback_allocator.d
@@ -66,7 +66,7 @@ public:
         }
         else
             assert(contains(point), "Attepmted to roll back to a point not in the allocator.");
-        while (!first.contains(point))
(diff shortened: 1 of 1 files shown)


WebFreak001:
sorry this is garbage, gonna add unittests and fix the current implementation of allocate rather

PR #382, opened 2020-02-09
Fix rollback_allocator implementation

WebFreak001:
Uses AlignedMallocator instead of Mallocator to guarantee memory alignment (and avoid "The memoriez" assert error)

Throws `out of memory` now whenever the allocator returns null.

Fixes rollback bug with over-allocation.

Fixes some null checks in code and asserts.

Added unittests for usage which the previous rollback allocator didn't pass.

(this fixes a lot of seemingly random crashes and bugs with big files)

### DIFF
--- a/src/dparse/rollback_allocator.d
+++ b/src/dparse/rollback_allocator.d
@@ -9,6 +9,9 @@ struct RollbackAllocator
 {
 public:
 
+    // must be multiple of 8
+    enum memoryAlignment = 16u;
+
     @disable this(this);
 
     ~this()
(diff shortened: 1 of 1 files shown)


WebFreak001:
@Hackerpilot I can haz review?

Added tests to test the functionality and otherwise the changes aren't so complex. The whole file is completely isolated so this is easy to review.

This fixes a bug in the allocation when having a large AST which can currently break my tools using this :)

src/dparse/rollback_allocator.d
```diff
@@ -9,6 +9,9 @@ struct RollbackAllocator
 {
 public:
 
+    // must be multiple of 8
+    enum memoryAlignment = 16u;
```
WebFreak001: changed to 16 to reflect the default of AlignedMallocator

PR #385, opened 2020-03-06
Fix error handling in invalid escape sequence

Hackerpilot:
Another problem uncovered while working on https://github.com/dlang-community/dfmt/issues/469


WebFreak001:
this (only handler assigning `Token.init`) breaks some libraries, eg. [D-Scanner](https://github.com/dlang-community/D-Scanner/issues/829) because it inserts empty tokens at 0 position, which is invalid and causes stuff like `t.line - 1` to underflow.

PR #389, opened 2020-04-17
Fixed unittest on x86

### LINKED ISSUE #388: Build failing on x86/32-bit.
Hello,

on x86 the build fails with:

```
 ldc2 -I=test_dparse@exe -I=. -I=.. -I=../src -I/usr/include/d/stdx-allocator -I/usr/include/d/mir-core -enable-color -unittest -of='test_dparse@exe/src_dparse_lexer.d.o' -c ../src/dparse/lexer.d
../src/dparse/lexer.d(2300): Error: function dparse.lexer.optimalBucketCount(uint size) is not callable using argument types (ulong)
../src/dparse/lexer.d(2300):        cannot pass argument 100000000000LU of type ulong to parameter uint size
```

Tested with libdparse 0.14 and 0.13.2.

WebFreak001:
fix #388

### DIFF
--- a/src/dparse/lexer.d
+++ b/src/dparse/lexer.d
@@ -1891,7 +1891,8 @@ unittest
 {
     assert(optimalBucketCount(1) == 2);
     assert(optimalBucketCount(9000 * 32) == 16384);
-    assert(optimalBucketCount(100_000_000_000UL) == 1 << 30);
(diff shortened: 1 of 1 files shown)


WebFreak001:
would maybe want to test x86 on travis, but not scope of this PR

PR #399, opened 2020-05-02
Fix 383 - Formatting AST including `@trusted` delegate produces inval…

MoonlightSentinel:
…id output

`do`was written unconditionally even though `hasDo` was false.

### DIFF
--- a/src/dparse/formatter.d
+++ b/src/dparse/formatter.d
@@ -2678,7 +2678,8 @@ class Formatter(Sink)
         {
             foreach (contract; functionContracts)
                 format(contract);
-            put("do");
+            if (specifiedFunctionBody.hasDo)
+                put("do");
             if (blockStatement)
                 format(blockStatement);
         }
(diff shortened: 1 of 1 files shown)


src/dparse/formatter.d
```diff
@@ -2678,7 +2678,8 @@ class Formatter(Sink)
         {
             foreach (contract; functionContracts)
                 format(contract);
-            put("do");
+            if (specifiedFunctionBody.hasDo)
+                put("do");
```
Geod24, line 2682: Looks like the line is not covered though ? And neither is `format(contract)`

MoonlightSentinel: Fixed.

MoonlightSentinel:
The second commit ensures proper newlines/indentation for do (and also in/out along the way).

Also added a bunch of tests.

Geod24:
Thanks! Looks like Travis had a spurious failure tho.

MoonlightSentinel:
Could you have a look @WebFreak001 ?

Geod24:
Can we have a merge & release then ? 👼

WebFreak001:
(stable) release will take a little because we have the token trivia changes in right now, which changed how ddoc of tokens was handled, which still needs a little more testing for backwards compatibility

PR #404, opened 2020-05-05
Fix #53 - Support GDC extended asm syntax

### LINKED ISSUE #53: Support GDC extended asm syntax
http://wiki.dlang.org/GDC/Using_GDC#Extended_Assembler

MoonlightSentinel:
This should already work for all "pure" GCC style `asm`-statements but probably needs to be revised against gdc (which is more permissive in certain aspects). Mostly looking for feedback regarding the generated AST.

~~WIP because this still needs some cleanup and mabye more test coverage.~~

TODO:
- [x] More reliable detection of GCC asm
- [x] Support arbitrary expressions as assmembler templates, not only strings
- [x] Issue deprecation warning for missing parantheses around operands

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -242,6 +242,9 @@ abstract class ASTVisitor
     /** */ void visit(const FunctionContract functionContract) { functionContract.accept(this); }
     /** */ void visit(const FunctionDeclaration functionDeclaration) { functionDeclaration.accept(this); }
     /** */ void visit(const FunctionLiteralExpression functionLiteralExpression) { functionLiteralExpression.accept(this); }
+    /** */ void visit(const GccAsmInstruction gccAsmInstruction) { gccAsmInstruction.accept(this); }
+    /** */ void visit(const GccAsmOperandList gccAsmOperands) { gccAsmOperands.accept(this); }
+    /** */ void visit(const GccAsmOperand gccAsmOperand) { gccAsmOperand.accept(this); }
     /** */ void visit(const GotoStatement gotoStatement) { gotoStatement.accept(this); }
     /** */ void visit(const IdentifierChain identifierChain) { identifierChain.accept(this); }
     /** */ void visit(const DeclaratorIdentifierList identifierList) { identifierList.accept(this); }
@@ -314,6 +317,7 @@ abstract class ASTVisitor
     /** */ void visit(const StaticDestructor staticDestructor) { staticDestructor.accept(this); }
     /** */ void visit(const StaticIfCondition staticIfCondition) { staticIfCondition.accept(this); }
     /** */ void visit(const StorageClass storageClass) { storageClass.accept(this); }
+    /** */ void visit(const StringLiteralList stringLiteralList) { stringLiteralList.accept(this); }
     /** */ void visit(const StructBody structBody) { structBody.accept(this); }
     /** */ void visit(const StructDeclaration structDeclaration) { structDeclaration.accept(this); }
     /** */ void visit(const StructInitializer structInitializer) { structInitializer.accept(this); }
@@ -800,9 +804,10 @@ final class AsmStatement : BaseNode
 {
     override void accept(ASTVisitor visitor) const
     {
-        mixin (visitIfNotNull!asmInstructions);
+        mixin (visitIfNotNull!(asmInstructions, gccAsmInstructions));
     }
     /** */ AsmInstruction[] asmInstructions;
+    /** */ GccAsmInstruction[] gccAsmInstructions;
     /** */ FunctionAttribute[] functionAttributes;
     mixin OpEquals;
 }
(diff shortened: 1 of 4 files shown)


Hackerpilot:
Sorry for the delay. I haven't forgotten about this.

MoonlightSentinel:
Don't worry about it, this just removes a lot of error messages from druntime and isn't blocking other work.

src/dparse/ast.d
```diff
@@ -800,9 +804,10 @@ final class AsmStatement : BaseNode
 {
     override void accept(ASTVisitor visitor) const
     {
-        mixin (visitIfNotNull!asmInstructions);
+        mixin (visitIfNotNull!(asmInstructions, gccAsmInstructions));
```
Hackerpilot, line 807: Looks like the `functionAttributes` were missing in the old version of this code. Not sure if that was an intentional omission.

src/dparse/parser.d
```diff
@@ -3556,6 +3593,128 @@ class Parser
...
+
+                if (skip(tok!":"))
+                {
+                    if (!currentIs(tok!":"))
+                        mixin(parseNodeQ!("node.registers", "StringLiteralList"));
+
+                    if (skip(tok!":"))
+                    {
+                        if (node.outputOperands.items.length)
+                            error("goto-labels only allowed without output operands!", false);
```
Hackerpilot, line 3647: I think there may be a `return null;` missing after this `error`.

src/dparse/parser.d
```diff
@@ -3556,6 +3593,128 @@ class Parser
...
+     *     ($(RULE gccAsmOperand) ($(LITERAL ',') $(RULE gccAsmOperand))* )?
+     * )
+     */
+    GccAsmOperandList parseGccAsmOperandList()
+    {
+        mixin(traceEnterAndExit!(__FUNCTION__));
+
+        // Disambiguate between an empty and a missing list
+        if (currentIsOneOf(tok!":", tok!";"))
+            return allocator.make!GccAsmOperandList();
```
Hackerpilot, line 3672: If I'm reading this logic here and the grammar correctly, this grammar rule can match an empty array of tokens?

Hackerpilot: If so, it may be better to make the `gccAsmOperandList` optional in whatever rule it's used, and then define this rule to always match a non-empty sequence of tokens.

src/dparse/parser.d
```diff
@@ -3556,6 +3593,128 @@ class Parser
...
+     *
+     * $(GRAMMAR $(RULEDEF gccAsmOperand)
+     *     ($(LITERAL '[') $(RULE identifier) $(LITERAL ']'))? $(RULE stringLiteral) $(LITERAL '(') $(RULE assignExpression) $(LITERAL ')')
+     * )
+     */
+    GccAsmOperand parseGccAsmOperand()
+    {
+        mixin(traceEnterAndExit!(__FUNCTION__));
+        if (currentIsOneOf(tok!":", tok!";"))
+            return null;
```
Hackerpilot, line 3688: Normally returning null indicates a parse error.

src/dparse/parser.d
```diff
@@ -3556,6 +3593,128 @@ class Parser
...
+        if (currentIsOneOf(tok!":", tok!";"))
+            return allocator.make!GccAsmOperandList();
+
+        return parseCommaSeparatedRule!(GccAsmOperandList, GccAsmOperand)();
+    }
+
+    /**
+     * Parses a GccAsmOperand
+     *
+     * $(GRAMMAR $(RULEDEF gccAsmOperand)
```
Hackerpilot, line 3680: There should be a colon at the end of this line. Same with several of the other grammar rules in the pull request.

src/dparse/parser.d
```diff
@@ -3556,6 +3593,128 @@ class Parser
...
+     *     ($(LITERAL '[') $(RULE identifier) $(LITERAL ']'))? $(RULE stringLiteral) $(LITERAL '(') $(RULE assignExpression) $(LITERAL ')')
```
Hackerpilot, line 3681: There should be a semicolon after the last part of the grammar rule. Same with some of the other grammar rules.

src/dparse/ast.d
MoonlightSentinel: No idea but I'll ad them to be consistent with the other AST nodes.

src/dparse/parser.d
MoonlightSentinel: Parsing the goto's anyway achieves better error reporting because it avoids parsing the remaining goto's as a new instruction.
But you're right, it should probably return null after that.

src/dparse/parser.d
MoonlightSentinel: Exactly. Moved this check to avoid the duplication for input and output operands.

src/dparse/parser.d
MoonlightSentinel: This method is called from `parseCommaSeparatedRule` which stops when this rule returns `null`. Or is there  a better way to do this?

Hackerpilot: I think that function works by parsing one of the comma-separated things, then checking to see if the current token is a comma, if it's not, then it stops parsing elements. It shouldn't be necessary for this to return null in the case that the list is done, because parseCommaSeparatedRule shouldn't try to call this function.

MoonlightSentinel: You're right, that wasn't necessary anymore. Removed.

src/dparse/parser.d
Hackerpilot: Is this change still planned?

MoonlightSentinel: Not really, I still dislike the resulting duplication. But i can make it anyway if you insist.

Hackerpilot: For the reasons that a) I really dislike the idea of empty AST nodes, and b) consistency withe the rest of the parser, I'm going to insist.

MoonlightSentinel: Done

MoonlightSentinel:
Thanks. A new release would be appreciated (once it's ready)

Hackerpilot:
https://github.com/dlang-community/libdparse/releases/tag/v0.15.0

PR #406, opened 2020-05-16
use CTFE functions in mixin(), and not templates

Jozef-r:
The duration spent to compile dparse as a static library is the same but this reduces memory usage.
This change will also begin being interesting when the new CTFE implementation will come.


Jozef-r:
If you want more conceptual explanations.  let's say you have a template `temp(T)`. Let the instance of `temp!int`. This instance is stored in a kind of hashmap so that the cost of making an instance for `int` only happens once.

The problem we are faced to here is that the cost of creating instances and making lookups in the template each time it's used again **is not worth**. It saves a very little, to the point that generating the code again does not make compiling slower. So instead, CTFE is used which has the side effect to polute less the symbol tables and reduce the memory used.

WebFreak001:
ok I did some small benchmarks how this affects compile time and memory usage with a simple `touch src/dparse/ast.d && time dub build` (always picking run with minimum time of 5 runs, LDC 1.20.1 on ivybridge):

Before: dub build time = 7.36s, peak memory = 831 MB
After: dub build time = 7.34s, peak memory = 826 MB

I don't know if that tiny improvement is really worth this change

Jozef-r:
Yes, i also see those 6 Megs in less using

```bash
for i in $(seq 1 5); do
  git checkout mixin-using-ctfe
  echo "LDC2 PR pass $i"
  /usr/bin/time -v dub build --build=release --compiler=ldc2 --force
  git checkout master
  echo "LDC2 MASTER pass $i"
  /usr/bin/time -v dub build --build=release --compiler=ldc2 --force
done

for i in $(seq 1 5); do
  git checkout mixin-using-ctfe
  echo "DMD PR pass $i"
  /usr/bin/time -v dub build --build=release --compiler=dmd --force
  git checkout master
  echo "DMD MASTER pass $i"
  /usr/bin/time -v dub build --build=release --compiler=dmd --force
done
```
As explained this is due to the fact that templates were not useful. They are used as a kind of cache, with the hope to reduce the cat (`~)` operations in CTFE, but this cache has for only effect to use more memory, although conceptually this was a good idea.

PR #407, opened 2020-05-19
Use the LPAREN and RPAREN macros in gcc asm grammar

Hackerpilot:
You can see the problem here: https://libdparse.dlang.io/grammar.html#gccAsmOperand


Geod24:
@Hackerpilot : Unrelated to this PR, but could you give a kick to the registry (https://code.dlang.org/packages/libdparse) ?

Hackerpilot:
No, because the part of the dub registry that sends password reset emails is about as fast as the part that scans for new release tags. I'll check in on this tomorrow.

PR #412, opened 2020-05-20
Fix, Named entities not lexed correctly

### LINKED ISSUE #405: Named entities not lexed correctly
test case

```d
module bug;

const a = "\&copy;";  // Declaration expected
const b = "\&trade;";  // Declaration expected
```

but those are valid declarations.

Jozef-r:
The strategy for better perfomances is to assume the name is correct as long as it's an ASCII alpha.
That will close #405.


src/dparse/lexer.d
```diff
@@ -1526,6 +1526,22 @@ private pure nothrow @safe:
...
     }
 
+    bool lexNamedEntity()
+    {
+        assert (range.bytes[range.index] == '&');
+
+        Token t;
+        range.popFront();
+        lexIdentifier(t);
+        if (t.type != tok!"identifier" || range.bytes[range.index] != ';')
```
WebFreak001, line 1536: could crash out-of-bounds here

src/dparse/lexer.d
```diff
@@ -1526,6 +1526,23 @@ private pure nothrow @safe:
...
 
+    bool lexNamedEntity()
+    {
+        assert (range.bytes[range.index] == '&');
+
+        Token t;
+        range.popFront();
+        lexIdentifier(t);
+        if (t.type != tok!"identifier" ||
+            !range.empty && range.bytes[range.index] != ';')
```
WebFreak001, line 1537: D-Scanner would complain that you need to add () around the && exp here to clarify precedence.

You should also either write out the `!range.empty` as the comparision it does or change the `range.bytes[...]` call to use the range interface. (`.front`) Mixing random access and range API calls is prone to errors for modifications

src/dparse/lexer.d
```diff
@@ -1526,6 +1526,23 @@ private pure nothrow @safe:
             index);
     }
 
+    bool lexNamedEntity()
+    {
+        assert (range.bytes[range.index] == '&');
```
WebFreak001, line 1531: I think it would be more D-style to write this kind of input as in contract. Because libdparse still supports older D versions in its travis file you need to do the full `in { assert(...); } do` code though.

src/dparse/lexer.d
```diff
@@ -2261,6 +2279,43 @@ unittest
...
 }
 
+unittest
+{
+    LexerConfig cf;
+    StringCache ca = StringCache(16);
+
+    {
+        auto l = DLexer(`"\&copy;"`, cf, &ca);
+        assert(l.front().type == tok!"stringLiteral");
```
WebFreak001, line 2289: can you test what the `.text` property of the tokens is?

WebFreak001: nvm this isn't part of this API

WebFreak001:
implemented with #420

PR #418, opened 2020-06-21
add dparse.strings utility to unescape strings

### LINKED ISSUE #417: StringBehavior is a lie
I tried
```d
unittest
{
        scope StringCache cache = StringCache(16);
        LexerConfig config;
        config.commentBehavior = CommentBehavior.noIntern;
        config.stringBehavior = StringBehavior.compiler; // interpret string literals
        config.whitespaceBehavior = WhitespaceBehavior.skip;
        config.fileName = "evaluate-string-stdin";
        auto input = q{"hello world"" foo\n bar"w" ok"};

        string[] expected = ["hello world", " foo\n bar", " ok"];
        int i = 0;
        foreach (t; DLexer(input, config, &cache))
                assert(t.text == expected[i++], t.text);
}
```

assuming StringBehavior.compiler would, as the documentation says it:
```
Do not include quote characters, process escape sequences
```

However now this test actually fails:
```
core.exception.AssertError@src/dparse/lexer.d(2364): "hello world"
```

You see that the string is passed as-is.

I have also not found any usage of StringBehavior at all except for token strings changing it temporarily for those. This is weird because StringBehavior is the only config thing that is fully documented (both enum decl and members) so I would expect that this thing should be implemented.

WebFreak001:
Deprecates the StringBehavior values, fixes #417 as alternative API

dparse.strings is fully tested & covered.

### DIFF
--- a/src/dparse/strings.d
+++ b/src/dparse/strings.d
@@ -132,8 +132,8 @@ enum InvalidEscapeAction
 	/// string like in the input string.
 	keep = 0,
 	/// Ignore and skip offending characters, drop them from the output. Named
-	/// character entities are still being included like keep as they are not
-	/// currently implemented.
+	/// character entities are still being included like $(LREF keep) as they
+    /// are not currently implemented.
 	skip,
 	/// Throw a ConvException on invalid escape sequences. Does not throw
 	/// anything on unknown named character entities as they are not currently
--- a/src/dparse/strings.d
+++ b/src/dparse/strings.d
@@ -439,7 +439,8 @@ private string unescapeDoubleQuotedContent(
 		final switch (invalidEscapeAction)
 		{
 		case InvalidEscapeAction.keep:
-			ret ~= input[start .. start = continueAt];
+			ret ~= input[start .. continueAt];
+			start = continueAt;
 			break;
 		case InvalidEscapeAction.skip:
 			start = continueAt;
@@ -535,9 +536,14 @@ private string unescapeDoubleQuotedContent(
 		case '&':
 			auto end = input.indexOf(';', escape + 2);
 			if (end == -1)
+			{
 				errorInvalidCharacter(input.length);
+			}
 			else
-				ret ~= input[escape .. start = end + 1];
+			{
+				ret ~= input[escape .. end + 1];
+				start = end + 1;
+			}
 			break;
 		default:
 			errorInvalidCharacter(escape + 1);
@@ -671,9 +677,7 @@ private string normalizeNewLines(string text)
 		else if (text[end] != '\r')
 			end += exoticLineBreakLength - 1;
 		start = end + 1;
-		end = text[start .. $].indexOfAny(nlCharacters);
-		if (end != -1)
-			end += start;
+		end = text.indexOfAny(nlCharacters, start);
 	}
 	ret ~= text[start .. $];
 	return ret.data;
--- a/src/dparse/strings.d
+++ b/src/dparse/strings.d
@@ -133,7 +133,7 @@ enum InvalidEscapeAction
 	keep = 0,
 	/// Ignore and skip offending characters, drop them from the output. Named
 	/// character entities are still being included like $(LREF keep) as they
-    /// are not currently implemented.
+	/// are not currently implemented.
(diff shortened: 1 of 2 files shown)


src/dparse/strings.d
```diff
@@ -0,0 +1,694 @@
...
+
+/// Defines different handler types what to do when invalid escape sequences are
+/// found inside $(LREF unescapeString).
+enum InvalidEscapeAction
+{
+	/// keep the backslash character as well as the escape characters in the
+	/// string like in the input string.
+	keep = 0,
+	/// Ignore and skip offending characters, drop them from the output. Named
+	/// character entities are still being included like keep as they are not
```
Hackerpilot, line 135: I think that it would be more clear to write this as "included like \`keep\` as they" to emphasize that the other enum member is being referenced.

src/dparse/strings.d
```diff
@@ -0,0 +1,694 @@
...
+	}
+
+	return ret.data;
+}
+
+private int parseHexChar(char c)
+in
+{
+	assert(c.isHexDigit);
+	assert('a' > 'A' && 'A' > '0'); // just checking that ASCII doesn't suddenly change
```
Hackerpilot, line 633: That's very unlikely.

src/dparse/strings.d
```diff
@@ -0,0 +1,694 @@
...
+/// Defines different handler types what to do when invalid escape sequences are
+/// found inside $(LREF unescapeString).
+enum InvalidEscapeAction
+{
+	/// keep the backslash character as well as the escape characters in the
+	/// string like in the input string.
+	keep = 0,
+	/// Ignore and skip offending characters, drop them from the output. Named
+	/// character entities are still being included like $(LREF keep) as they
+    /// are not currently implemented.
```
Hackerpilot, line 136: The indentation here is not consistent.

src/dparse/strings.d
```diff
@@ -0,0 +1,694 @@
...
+	assertThrown!ConvException(unescapeString(`x"AG"`));
+	assert(unescapeString!(InvalidEscapeAction.keep)(`x"AG"`) == "AG");
+	assert(unescapeString!(InvalidEscapeAction.skip)(`x"AG"`) == "");
+
+	assertThrown!ConvException(unescapeString(`x"A"`));
+	assert(unescapeString!(InvalidEscapeAction.keep)(`x"A"`) == "A");
+	assert(unescapeString!(InvalidEscapeAction.skip)(`x"A"`) == "");
+}
+
+private string unescapeDoubleQuotedContent(
```
Hackerpilot, line 400: It's going to take me a bit to get through this. I'll come back to it tomorrow.

src/dparse/strings.d
WebFreak001: ah the github online editor sucks, but it's so much more convenient to open

src/dparse/strings.d
WebFreak001: well it does make the intent in the function body clearer (not having maximum checks for the character values) and shows my assumption about the characters, which maybe isn't clear to someone or someone doesn't know the ASCII ordering out the top of their head.

Both because this is an assert (in an `in` block too) and because this is evaluatable at compile time, this should be removed (at least in release mode), so there is no overhead to this.

src/dparse/strings.d
```diff
@@ -0,0 +1,694 @@
...
+			return true;
+		}
+	}
+
+	void errorInvalidCharacter(size_t continueAt)
+	{
+		final switch (invalidEscapeAction)
+		{
+		case InvalidEscapeAction.keep:
+			ret ~= input[start .. start = continueAt];
```
Hackerpilot, line 442: Can this assignment be moved somewhere else? I'm guessing that it evaluates the left of the `..` first, then does the assignment, and then evaluates the `start` on the right side, but that's not easy to read.

WebFreak001:
would you put this in a 0.15.2 or in a 0.16.0? I would personally tend to put this in 0.15.2 to not need to upgrade the other tools again once they upgraded to >=0.15.0 because this is only additions and deprecations for the future

PR #420, opened 2020-06-28
Fix #405: Named entities not lexed correctly

### LINKED ISSUE #405: Named entities not lexed correctly

MoonlightSentinel:
Rebase of #412  with the requested changes.

### DIFF
--- a/src/dparse/lexer.d
+++ b/src/dparse/lexer.d
@@ -1537,6 +1537,22 @@ private pure nothrow @safe:
             index);
     }
 
+    bool lexNamedEntity()
+    in { assert (range.bytes[range.index] == '&'); }
+    do
+    {
+        Token t;
+        range.popFront();
+        lexIdentifier(t, true);
+        if (t.type != tok!"identifier" || range.empty || range.bytes[range.index] != ';')
+        {
+            error("Error: invalid named character entity");
+            return false;
+        }
+        range.popFront();
+        return true;
+    }
+
     bool lexEscapeSequence()
     {
         range.popFront();
@@ -1547,6 +1563,7 @@ private pure nothrow @safe:
         }
         switch (range.bytes[range.index])
         {
+        case '&': return lexNamedEntity();
         case '\'':
         case '"':
         case '?':
@@ -1699,11 +1716,14 @@ private pure nothrow @safe:
         }
     }
 
-    void lexIdentifier(ref Token token) @trusted
+    void lexIdentifier(ref Token token, const bool silent = false) @trusted
     {
         mixin (tokenStart);
+
         if (isSeparating(0))
         {
+            if (silent) return;
+
             error("Invalid identifier");
             range.popFront();
         }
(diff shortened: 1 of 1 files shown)


Geod24:
You mean rebase of #412 to fix #405 I assume.

src/dparse/lexer.d
```diff
@@ -1537,6 +1537,23 @@ private pure nothrow @safe:
...
+    in { assert (range.bytes[range.index] == '&'); }
+    do
+    {
+        Token t;
+        range.popFront();
+        lexIdentifier(t);
+        if (t.type != tok!"identifier" ||
+            (!range.empty && range.bytes[range.index] != ';'))
+        {
+            error("Error: invalid named character entity");
```
Geod24, line 1550: There's no test for this, is there ?

MoonlightSentinel: Extended the test to check the actual messages.

src/dparse/lexer.d
```diff
@@ -2272,6 +2290,44 @@ unittest
...
+    }
+    {
+        auto l = DLexer(`"\&trade;\&urcorn"`, cf, &ca);
+        assert(l.front().type == tok!"");
+        assert(l.messages == [ Msg(1, 18, "Error: invalid named character entity", true) ]);
+    }
+    {
+        auto l = DLexer(`"\&"`, cf, &ca);
+        assert(l.front().type == tok!"");
+        assert(l.messages == [ Msg(1, 4, "Invalid identifier", true), Msg(1, 6, "Error: unterminated string literal", true) ]);
```
MoonlightSentinel, line 2322: Not sure if this one is acceptable...

MoonlightSentinel: The second commit adds a potential workaround.

MoonlightSentinel:
Ping

src/dparse/lexer.d
```diff
@@ -1537,6 +1537,23 @@ private pure nothrow @safe:
...
+        Token t;
+        range.popFront();
+        lexIdentifier(t, true);
+        if (t.type != tok!"identifier" ||
+            (!range.empty && range.bytes[range.index] != ';'))
+        {
+            error("Error: invalid named character entity");
+            return false;
+        }
+        range.popFront();
```
WebFreak001, line 1553: I believe this can RangeError if the tokens end with `&` `identifier`, because we consumed the ampersand and the identifier, but the if invalid check didn't trigger because range.empty was true

For the same effect but a more sturdy check I would rewrite the if to
```d
if (t.type != tok!"identifier" || range.empty || range.bytes[range.index] != ';')
```

MoonlightSentinel: Fixed. Also added a test case for `"\&0`

WebFreak001: that test case wouldn't match anyway because it already fails on the identifier, make the test case use an identifier instead

MoonlightSentinel: Done

Issue #423, opened 2020-07-25
implement throw attribute


WebFreak001:
See https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1029.md which has been now accepted

Hackerpilot:
Fixed by #424.

Issue #143, opened 2017-06-13
Add AST dumps to the test suite


wilzbach:
- [ ] add a simple XML output (like Dscanner, or maybe even use dscanner)
- [ ] add AST dumps to fix results and improve code coverage

Related to #129

Hackerpilot:
To check that AST nodes are being generated correctly we will need to make a test system that can be given a bunch of source code, and then corresponding ASTs, or queries against the generated AST, that should match.

One way to do this is to create some sort of AST query testing tool and then make a set of unit tests in the code.

Another way, which I like because it's lazier, is to move the `dscanner.astprinter` module from the D-Scanner project into libdparse, and then create a small testing program that will read in code and the output the AST in XML format. We can then re-use the tree querying functionality of the xmllint program by writing a script that takes test files and a set of XPath expressions to run on it. We'd have to make sure that the CI systems are configured to install the xmllint tool, but I don't think that should be too difficult.

wilzbach:
> Another way, which I like because it's lazier, is to move the dscanner.astprinter module from the D-Scanner project into libdparse,

I would go/vote for this one :+1:

PR #424, opened 2020-07-27
Support 'throw' as a function attribute

Hackerpilot:
See https://github.com/dlang-community/libdparse/issues/423

### DIFF
--- a/test/pass_files/throwattribute.d
+++ b/test/pass_files/throwattribute.d
@@ -1,0 +1,10 @@
+void doStuff() throw {
+	throw void doThings()
+	{
+
+	}
+	throw new Exception("OH NO!");
+}
+
+throw:
+	extern void throwsSomething();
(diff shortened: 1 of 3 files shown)


Hackerpilot:
The example file produces the correct AST, but this could probably use a bit more testing.

Hackerpilot:
After thinking about it a bit more, I think it's better to get this change merged in so that the various tools that depend on libdparse can support this syntax when the next compiler release comes out. Testing that the parser produces the correct AST (See https://github.com/dlang-community/libdparse/issues/143) is a much bigger issue and makes sense to do separately from this.

test/pass_files/throwattribute.d
```diff
@@ -0,0 +1,10 @@
+void doStuff() throw {
+	throw void doThings()
```
AndrejMitrovic, line 2: What about a fail case for `void foo (throw int x)`? Or does it not matter / it's already handled?

Hackerpilot: It doesn't parse.

Hackerpilot: And, actually, putting a test case for that into the fail_files directory was really easy, so I did it.

PR #425, opened 2020-07-31
Add an XPath-based testing system for the AST generated by the parser

Hackerpilot:
As discussed in https://github.com/dlang-community/libdparse/issues/143, this adds an XPath-based testing system for checking that the AST generated by the parser is what we expect. Right now there's just one test, which resembles some code that triggered a soon-to-be-fixed DCD issue.

* The XML AST printer was copied from D-Scanner
* A small program in the test directory is added that does a D->XML translation
* The Travis CI configuration now specifies that the `xmllint` tool must be installed.
* The test script loops over all files in the new ast_checks folder
  * For each of these files it looks for a corresponding txt file
  * Each line of this txt file is treated as an XPath expression that `xmllint` then checks against the AST generated above
  * A summary of failing lines is printed to the console if there are any.

This is only designed to allow tests for expressions that match, i.e. you cannot use this system to check that an XPath expression does not match the AST, but I don't think that we'll need that.

I think this should be enough to finally close https://github.com/dlang-community/libdparse/issues/143


Hackerpilot:
Irony: Adding the AST checks causes the code coverage to drop, and thus our systems tell us that the library is less well tested.

Hackerpilot:
All right. `--DRT-testmode=run-main` was missing from the coverage tests, which has caused the coverage to be wrong for a while. Adding this gets back the behavior that we used to have with older versions of DMD.

PR #427, opened 2020-08-02
Fix 426: Storage classes for C style variadic functions

MoonlightSentinel:


src/dparse/parser.d
```diff
@@ -5288,8 +5294,27 @@ class Parser
...
+                if (!currentIs(tok!"..."))
+                    return null;
+
+                // Reset to the beginning of the current parameters
+                index = startIdx;
+                allocator.rollback(cp);
+
+                node.hasVarargs = true;
+                mixin(parseNodeQ!(`node.varargsAttributes`, `VariadicArgumentsAttributes`));
+                expect(tok!"...");
```
Hackerpilot, line 5315: I think that this should be `if (expect(tok!"...") is null) return null;`, as the parser functions should return null if there is an error. I know that not a lot of the other parse functions work this way right now, but that's a but that needs to be fixed some other time.

src/dparse/parser.d
```diff
@@ -5301,6 +5326,54 @@ class Parser
...
+        ParameterAttribute[] buffer;
+        ownArray(buffer, attributes);
+        return buffer;
+    }
+
+    /**
+     * Parses an attribute of C-style variadic parameters.
+     *
+     * $(GRAMMAR $(RULEDEF variadicArgumentsAttribute):
+     *       $(RULE variadicArgumentsAttribute)+
```
Hackerpilot, line 5356: This looks copied and pasted.

src/dparse/parser.d
```diff
@@ -5301,6 +5326,54 @@ class Parser
...
+
+        if (!currentIsOneOf(tok!"const", tok!"immutable", tok!"shared", tok!"scope", tok!"return"))
+        {
+            error("`const`, `immutable`, `shared`, `scope`, `return` expected");
+            return null;
+        }
+
+        node.idType = current.type;
+        advance();
+        node.tokens = tokens[startIndex .. index];
```
Hackerpilot, line 5373: Can't this array only ever have a single element in it? There's only a single call to `advance` here and there aren't any loops in this function.

WebFreak001: `.tokens` comes from the base AST class which is a range to the tokens some AST node is constructed from, so it's always a range of tokens (and this is the boilerplate similar to all other functions I assume)

MoonlightSentinel: Tried to follow the usual style, could also write it this way if preferred.

```suggestion
        node.tokens = tokens[index - 1 .. index];
```

src/dparse/parser.d
MoonlightSentinel: Fixed

src/dparse/parser.d
MoonlightSentinel: Fixed

src/dparse/parser.d
```diff
@@ -5301,6 +5326,58 @@ class Parser
...
+        ParameterAttribute[] buffer;
+        ownArray(buffer, attributes);
+        return buffer;
+    }
+
+    /**
+     * Parses an attribute of C-style variadic parameters.
+     *
+     * $(GRAMMAR $(RULEDEF variadicArgumentsAttribute):
+      *      $(LITERAL 'const')
```
Hackerpilot, line 5356: Is there an extra space on this line?

MoonlightSentinel: Not anymore

PR #429, opened 2020-08-11
Fix #428

### LINKED ISSUE #428: ConditionalDeclaration.trueStyle not set correctly for debug conditions
Example:
```d
debug:
const a = 1;
```
The `trueStyle` field of the resulting AST node is `single` when it should clearly be `colon`.

Hackerpilot:
This fixes a few logic problems in the handling of the `trueStyle` and `falseStyle` fields. The worst of these is that the old code called `advance` to skip over a token before then checking what the token was by looking at the new current token. It also had logic to set the field to `DeclarationListStyle.single` inside of an `if` that checked that there was a `{` or `:`, which makes no sense.

### DIFF
--- a/src/dparse/astprinter.d
+++ b/src/dparse/astprinter.d
@@ -218,13 +218,13 @@ class XMLPrinter : ASTVisitor
 	{
 		output.writeln("<conditionalDeclaration>");
 		visit(conditionalDeclaration.compileCondition);
-		output.writeln("<trueDeclarations>");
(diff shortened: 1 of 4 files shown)


WebFreak001:
want to add tests or merge as is?

Hackerpilot:
I think that our options are
1. Merge as-is
2. Or
 a. merge https://github.com/dlang-community/libdparse/pull/425 first
 b. change the XML output for the node to actually say what kind of compile condition it is, so that the XPATH can check it
 c. write a test

Hackerpilot:
I merged the AST testing system and now this has a regression test.

PR #436, opened 2020-09-24
Support parsing named function and template arguments.

Hackerpilot:
https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1030.md

The DIP mentions changing the definition of the `ArgumentList` rule, but that would cause problems in places such as case statements, mixins, index expression, and several others. I'll need to make sure that the pull request to the grammar spec is correct when it happens. A few other rules should be changed to use the new `NamedArgumenList`, and others need to continue using `ArgumentList`.


WebFreak001:
thanks! This is gonna be quite an exciting change.

Further things on the roadmap:
- dfmt should support these as multi-line init when a line gets too large
- DCD should ideally auto-complete available keys, but can probably already reuse the calltips for that

Will review and adjust this PR

PR #441, opened 2021-05-06
Make parser recognize AliasAssign

RazvanN7:
As per: https://github.com/dlang/dmd/pull/11846

A few notes:
 - I have added a new class `AliasAssign` that is viewed as a declaration so that it could appear in the same context as declarations. This might be a bit hacky, but it seemed like the easiest way to implement this.
 - The `AliasAssign` class is just a wrapper for `AliasInitializer`. I considered using an `AliasInitializer` as that would make the diff smaller, however, that comes with 2 drawbacks: comments cannot be attached to `AliasInitializer`s and it uses `AliasInitializer` for a purposes that it wasn't designed for. If the maintainers come to the conclusion that it is better to use `AliasInitializer`, I am open to amend the PR.
 - AliasAssignments should be present only in template declaration contexts, therefore I have added another default parameter to `parseDeclaration`, so that is signaled to the declaration.

This is currently blocking some PRs in phobos, so if you merge, please also release a new tag so that we can unblock those PRs.

Cheers,
RazvanN

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -152,6 +152,7 @@ abstract class ASTVisitor
 
     /** */ void visit(const AddExpression addExpression) { addExpression.accept(this); }
     /** */ void visit(const AliasDeclaration aliasDeclaration) { aliasDeclaration.accept(this); }
+    /** */ void visit(const AliasAssign aliasAssign) { aliasAssign.accept(this); }
     /** */ void visit(const AliasInitializer aliasInitializer) { aliasInitializer.accept(this); }
     /** */ void visit(const AliasThisDeclaration aliasThisDeclaration) { aliasThisDeclaration.accept(this); }
     /** */ void visit(const AlignAttribute alignAttribute) { alignAttribute.accept(this); }
@@ -485,6 +486,19 @@ final class AliasDeclaration : BaseNode
     /** */ string comment;
     /** */ Parameters parameters;
     /** */ MemberFunctionAttribute[] memberFunctionAttributes;
+}
+
+///
+final class AliasAssign : BaseNode
+{
+    override void accept(ASTVisitor visitor) const
+    {
+        mixin (visitIfNotNull!(identifier, type));
+    }
+    mixin OpEquals;
+    /** */ Token identifier;
+    /** */ Type type;
+    /** */ string comment;
 }
 
 ///
@@ -1279,7 +1293,7 @@ final class Declaration : BaseNode
     private import std.variant:Algebraic;
     private import std.typetuple:TypeTuple;
 
-    alias DeclarationTypes = TypeTuple!(AliasDeclaration, AliasThisDeclaration,
+    alias DeclarationTypes = TypeTuple!(AliasDeclaration, AliasAssign, AliasThisDeclaration,
         AnonymousEnumDeclaration, AttributeDeclaration,
         ClassDeclaration, ConditionalDeclaration, Constructor, DebugSpecification,
         Destructor, EnumDeclaration, EponymousTemplateDeclaration,
@@ -1302,6 +1316,7 @@ final class Declaration : BaseNode
     /** */ Declaration[] declarations;
 
     mixin(generateProperty("AliasDeclaration", "aliasDeclaration"));
+    mixin(generateProperty("AliasAssign", "aliasAssign"));
     mixin(generateProperty("AliasThisDeclaration", "aliasThisDeclaration"));
     mixin(generateProperty("AnonymousEnumDeclaration", "anonymousEnumDeclaration"));
     mixin(generateProperty("AttributeDeclaration", "attributeDeclaration"));
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -189,6 +189,25 @@ class Parser
     }
 
     /**
+     * Parses an AliasAssign.
+     *
+     * $(GRAMMAR $(RULEDEF aliasAssign):
+     *       $(LITERAL Identifier) $(LITERAL '=') $(RULE type)
+     */
+    AliasAssign parseAliasAssign()
+    {
+        mixin(traceEnterAndExit!(__FUNCTION__));
+        auto startIndex = index;
+        auto node = allocator.make!AliasAssign;
+        node.comment = comment;
+        comment = null;
+        mixin(tokenCheck!(`node.identifier`, "identifier"));
+        mixin(tokenCheck!"=");
+        mixin(parseNodeQ!(`node.type`, `Type`));
+        return attachCommentFromSemicolon(node, startIndex);
+    }
+
+    /**
      * Parses an AliasInitializer.
      *
      * $(GRAMMAR $(RULEDEF aliasInitializer):
@@ -1809,7 +1828,7 @@ class Parser
      *     | $(RULE compileCondition) $(LITERAL ':') $(RULE declaration)+ $(LITERAL 'else') $(LITERAL ':') $(RULE declaration)*
      *     ;)
      */
-    ConditionalDeclaration parseConditionalDeclaration(bool strict)
+    ConditionalDeclaration parseConditionalDeclaration(bool strict, bool inTemplateDeclaration = false)
     {
         mixin(traceEnterAndExit!(__FUNCTION__));
         auto startIndex = index;
@@ -1825,7 +1844,7 @@ class Parser
             while (moreTokens() && !currentIs(tok!"}") && !currentIs(tok!"else"))
             {
                 immutable c = allocator.setCheckpoint();
-                if (!trueDeclarations.put(parseDeclaration(strict, true)))
+                if (!trueDeclarations.put(parseDeclaration(strict, true, inTemplateDeclaration)))
                 {
                     allocator.rollback(c);
                     return null;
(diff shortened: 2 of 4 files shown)


RazvanN7:
Test suite is failing with "All tests have passed".

src/dparse/ast.d
```diff
@@ -487,6 +488,18 @@ final class AliasDeclaration : BaseNode
...
+///
+final class AliasAssign : BaseNode
+{
+    override void accept(ASTVisitor visitor) const
+    {
+        mixin (visitIfNotNull!(initializer));
+    }
+    mixin OpEquals;
+    /** */ AliasInitializer initializer;
+    /** */ string comment;
```
MoonlightSentinel, line 500: Shouldn't this node contain the identifier of the alias that will be assigned?

src/dparse/parser.d
```diff
@@ -188,6 +188,24 @@ class Parser
         return attachCommentFromSemicolon(node, startIndex);
     }
 
+    /**
+     * Parses an AliasAssign.
+     *
+     * $(GRAMMAR $(RULEDEF aliasDeclaration):
```
MoonlightSentinel, line 194:
```suggestion
     * $(GRAMMAR $(RULEDEF aliasAssign):
```

src/dparse/parser.d
```diff
@@ -6684,7 +6711,7 @@ class Parser
         while (moreTokens() && !currentIs(tok!"}"))
         {
             immutable c = allocator.setCheckpoint();
-            if (!declarations.put(parseDeclaration(true, true)))
+            if (!declarations.put(parseDeclaration(true, true, true)))
```
MoonlightSentinel, line 6714:
```suggestion
            if (!declarations.put(parseDeclaration(true, true, false, true)))
```

src/dparse/ast.d
RazvanN7: The `AliasInitializer` contains both the identifier and the initializer.

src/dparse/parser.d
RazvanN7: I don't understand, `parseDeclaration` has only 3 parameters.

src/dparse/parser.d
RazvanN7: Fixed.

src/dparse/parser.d
MoonlightSentinel: Nevermind

src/dparse/parser.d
```diff
@@ -188,6 +188,24 @@ class Parser
...
+     *
+     * $(GRAMMAR $(RULEDEF aliasAssign):
+     *       $(RULE aliasInitializer)
+     */
+    AliasAssign parseAliasAssign()
+    {
+        mixin(traceEnterAndExit!(__FUNCTION__));
+        auto startIndex = index;
+        auto node = allocator.make!AliasAssign;
+        node.initializer = parseAliasInitializer();
```
WebFreak001, line 202: AliasInitializer allows adding template parameters, are these really supported by alias assign? (as this is a public function that may be called freely - and for good code style)

test/pass_files/alias_assign.d
```diff
@@ -0,0 +1,14 @@
...
+     struct B
+     {
+          A t;   
+     }
+}
+
+void fun()
+{
+    auto b = T!float.B();
+}
```
WebFreak001, line 14: really weak test, add more complex nesting, conditional compilation, static foreach, etc. what would be used with this feature.

WebFreak001 (changes requested):
thanks for the PR! Got some points that should be changed

src/dparse/parser.d
```diff
@@ -188,6 +188,24 @@ class Parser
         return attachCommentFromSemicolon(node, startIndex);
     }
 
+    /**
+     * Parses an AliasAssign.
+     *
+     * $(GRAMMAR $(RULEDEF aliasAssign):
+     *       $(RULE aliasInitializer)
```
WebFreak001, line 195: this documentation says it would be syntactically valid to use templateParameters in an assignment. As it's never the case that it would actually be valid code I think it would be better to define a proper grammar here.

test/pass_files/alias_assign.d
```diff
@@ -0,0 +1,14 @@
+template T(X)
+{
+     alias A = int;
```
maxhaton, line 3: To add to what WebFreak has said, and being more specific, the code DScanner is blocking should be used as tests e.g. nu-staticMap etc.

src/dparse/parser.d
```diff
@@ -2100,7 +2120,7 @@ class Parser
      *     | $(RULE versionSpecification)
      *     ;)
      */
-    Declaration parseDeclaration(bool strict = false, bool mustBeDeclaration = false)
+    Declaration parseDeclaration(bool strict = false, bool mustBeDeclaration = false, bool inTemplateDeclaration = false)
```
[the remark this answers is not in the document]
RazvanN7: As far as I could tell from the implementation [1], AliasAssign is allowed only inside `template A(..){...}` declarations.

I understand your point, however, it is not possible to deduce whether the `static if` is at function level or at template level during parsing. In dmd, this is checked during semantic analysis, so I assume that the solution for libdparse is to create an AliasAssignDeclaration every time we encounter an assignment that occurs in non-function contexts. ParseDeclaration seems to be called at function scope level also. Any ideas on how I could differentiate between assignments at function level and assignments non-function levels?
 

[1] https://github.com/dlang/dmd/pull/11846/files?file-filters%5B%5D=.d&file-filters%5B%5D=.h#diff-90e6074482976dda59fff5b12d8aa29f1af3012b5db226a827e8e304b953ec10R6774

src/dparse/parser.d
RazvanN7: Nope. I can implement it properly, but I thought I would just reuse as much code as possible to make the diff smaller.

test/pass_files/alias_assign.d
RazvanN7: Fair point. I just wanted to get this out there as fast as possible as to get feedback.

src/dparse/parser.d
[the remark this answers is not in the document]
WebFreak001: I think you want to nest the inTemplateDeclaration parameter into all calls that should keep it, ie. here the parseConditionalDeclaration function and all the other `parse*Declaration` functions you want to support.

I personally think it would be more pretty doing this with some kind of parse state struct that is passed as argument in all related functions though if this is expected to expand in the future to more edge cases.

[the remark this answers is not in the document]
RazvanN7: I'm not sure this will work. Consider this case:

```d
template X(T)
{
    alias X = int;
    static if (cond)
        X = float;            // this one is currently accepted by dmd
    void fun()
    {
         X = T;               // while this one is not because the assignment
                                 // does not have the same parent as the declaration of X
    }
}
```

This is not going to get solved by passing the `inTemplate` boolean. Or maybe it could work, but that would mean inserting a primitive form of semantic analysis. Right now, libdparse issues an error when
assignments occur in non-function contexts. If we want to keep things in parser territory, we need to relax that condition and to create AliasAssignDeclarations if an assignment occurs in non-function context. Of course that will allow for code that is not an alias assignment to pass parsing, but that is fine.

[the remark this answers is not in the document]
WebFreak001: the parseFunctionDeclaration body would reset back the boolean value that you are in a template, (implicitly by simply not passing it down as parameter or for a struct explicitly) you will want to do that with anything adding scope.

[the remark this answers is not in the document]
RazvanN7: hmm, I'll try that out. Thanks for your assistance!

RazvanN7:
@WebFreak001 @maxhaton I gave it another pass. Kindly asking for a review

RazvanN7:
@WebFreak001 I agree that keeping all the state in one place is easier to maintain, however, at this point that might be an overkill. Anyway, I would rather merge this ASAP to unblock the phobos PRs and refactor later if needed.

WebFreak001:
travis fail unrelated

RazvanN7:
There seems to be a dependency error on stdx.allocator:

```
../subprojects/stdx-allocator/source/stdx/allocator/internal.d(722): Error: static assert:  `!__traits(compiles, emplace(& s2, 1))` is false
```

How do we get passed that? Did I mention that this PR is needed to unblock phobos :-" ?

TungstenHeart:
Try to change the version operator `'>= 2.77'` here

https://github.com/dlang-community/libdparse/blob/6f0893f6b07ec561d382f26c81e74f7788056828/meson.build#L10

to `'==2.77.5'` or anything (correct, bc I'm not sure of the syntax) that prevent to select the 3.0.0 branch. All dlang-community tools use the git 2.77.z branch for a similar reason.

If that does not work then I suggest to diable the Meson-driven tests. People who care will eventually come to fix the problem.

RazvanN7:
@WebFreak001 I have disabled the meson build as it seemed redundant. Once the dependency issues for meson are fixed, the tests can be un-commented. Is that acceptable?

TungstenHeart (approved):

PR #443, opened 2021-05-13
Update stdx-allocator.wrap

TungstenHeart:
The meson.build version for the v2.77.5 tag is 2.77.2, use a git hash instead

### DIFF
--- a/subprojects/stdx-allocator.wrap
+++ b/subprojects/stdx-allocator.wrap
@@ -1,4 +1,4 @@
 [wrap-git]
 directory = stdx-allocator
 url       = https://github.com/dlang-community/stdx-allocator.git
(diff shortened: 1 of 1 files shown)


TungstenHeart:
See the mismatch here : https://github.com/dlang-community/stdx-allocator/blob/v2.77.5/meson.build#L4.

Otherwise The proper way is to update the version to 2.77.6 in the 2.77.z branch then only after that push a git tag 2.77.6 on this commit.

TungstenHeart:
Then after the stfx-alloc fix this new version must also be used here 

https://github.com/dlang-community/libdparse/blob/6f0893f6b07ec561d382f26c81e74f7788056828/subprojects/stdx-allocator.wrap#L4

PR #449, opened 2021-09-03
lexer: avoid unnecessary allocations on compile-time arrays

ljmf00:
Signed-off-by: Luís Ferreira <contact@lsferreira.net>

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -49,7 +49,10 @@ jobs:
       - name: Install GDC
         if: ${{ matrix.compiler == 'gdc-latest' }}
         run: |
-          sudo apt-get install gdc gdmd -y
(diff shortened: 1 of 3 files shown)


ljmf00:
There is no CI?

rikkimax:
> There is no CI?

Doesn't look like it, hasn't been swapped over from Travis to GH actions.

WebFreak001:
seems to fail with GDC

ljmf00:
> seems to fail with GDC

I think this should be accepted by the compiler. I can add `.idup` tho, but seems a compiler bug? CC @ibuclaw

WebFreak001:
DCD build failure is due to [issues.dlang.org/show_bug.cgi?id=23874](https://issues.dlang.org/show_bug.cgi?id=23874)

PR #451, opened 2022-01-12
Implement `ThrowExpression`'s proposed by DIP 1034

MoonlightSentinel:
`ThrowExpression`'s replace `ThrowStatement`'s which may appear nested into any other declaration.

- dlang/dmd#13162
- dlang/dlang.org#3164

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -344,7 +344,7 @@ abstract class ASTVisitor
     /** */ void visit(const TemplateValueParameter templateValueParameter) { templateValueParameter.accept(this); }
     /** */ void visit(const TemplateValueParameterDefault templateValueParameterDefault) { templateValueParameterDefault.accept(this); }
     /** */ void visit(const TernaryExpression ternaryExpression) { ternaryExpression.accept(this); }
-    /** */ void visit(const ThrowStatement throwStatement) { throwStatement.accept(this); }
+    /** */ void visit(const ThrowExpression throwExpression) { throwExpression.accept(this); }
     /** */ void visit(const Token) { }
     /** */ void visit(const TraitsExpression traitsExpression) { traitsExpression.accept(this); }
     /** */ void visit(const TryStatement tryStatement) { tryStatement.accept(this); }
@@ -2376,7 +2376,7 @@ final class StatementNoCaseNoDefault : BaseNode
             whileStatement, doStatement, forStatement, foreachStatement,
             switchStatement, finalSwitchStatement, continueStatement,
             breakStatement, returnStatement, gotoStatement, withStatement,
-            synchronizedStatement, tryStatement, throwStatement,
+            synchronizedStatement, tryStatement,
             scopeGuardStatement, asmStatement, pragmaStatement,
             conditionalStatement, staticAssertStatement, versionSpecification,
             debugSpecification, expressionStatement, staticForeachStatement));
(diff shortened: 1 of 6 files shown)


src/dparse/ast.d
```diff
@@ -3188,6 +3201,7 @@ final class TernaryExpression : ExpressionNode
 }
 
 ///
+// deprecated("To be replaced by ExpressionStatement + ThrowExpression")
```
MoonlightSentinel, line 3204: Not sure what's the proper deprecation process is. 
Couldn't mark `ThrowStatement` as deprecated due to [this bug](https://issues.dlang.org/show_bug.cgi?id=22668).

WebFreak001: I think it might be best to remove/rename it completely (causing compiler errors) and have the new behavior already and do a version bump. Currently as I see it this change would mean ThrowStatement wouldn't be used at all anymore after some point, causing old code to still compile but no longer work as expected.

MoonlightSentinel: > [...]  ThrowStatement wouldn't be used at all anymore after some point, causing old code to still compile but no longer work as expected.

The idea was to remove `ThrowStatement` once it's no longer parsed. Any program that has been adapted to  `ThrowExpression` would keep working as expected and could simply remove the `ThrowStatement`-related code.

But outright removing it now should also be fine - assuming that users can likely transition by replacing  `ThrowStatement`. with `ThrowExpression`.

WebFreak001: if they dependend on ThrowStatement: (in a visitor)
- if we keep it and also add ThrowExpression, the visit function is likely no longer going to be called as expected anymore or have some changes in accessed member properties (potentially causing silent breakage)
- if we remove it and move all the relevant properties into ThrowExpression, the code will no longer compile and they can see that they need to update their code (loud breakage, no runnable wrong code)

So I think removing it is the better choice.

sadly I don't think we can alias ThrowStatement to some broken type to cause additional error messages, which would be optimal in guiding users how to upgrade.

src/dparse/ast.d
```diff
@@ -3187,16 +3198,8 @@ final class TernaryExpression : ExpressionNode
...
-{
-    override void accept(ASTVisitor visitor) const
-    {
-        mixin (visitIfNotNull!(expression));
-    }
-    /** */ Expression expression;
-    mixin OpEquals;
-}
+deprecated("Replaced by ExpressionStatement + ThrowExpression")
+final class ThrowStatement;
```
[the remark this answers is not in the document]
MoonlightSentinel: Seems like LDC is the only compiler to emit the `ClassInfo` reference.
Fixed by using a `struct` instead of a `class`

src/dparse/ast.d
```diff
@@ -3954,3 +3959,75 @@ unittest // Support GCC-sytle asm statements
...
+    },
+    (throwExp, throwable)
+    {
+        auto id = throwable.primaryExpression
+                            .identifierOrTemplateInstance
+                            .identifier;
+
+        assert(id.text == "someThrowable");
+    });
+}
```
WebFreak001, line 4033: I think you can replace these tests with better readable ones in https://github.com/dlang-community/libdparse/tree/master/test/ast_checks using XML paths and just plain .d files

MoonlightSentinel: Done

src/dparse/ast.d
MoonlightSentinel: The old `ThrowStatement` is now an alias to `ThrowExpression` and the field names are now identical. So existing code referencing `ThrowStatement` should keep working with a deprecation message.

MoonlightSentinel:
Ping. The implementation & documentation was merged upstream

PR #439, opened 2021-03-28
Support parsing of shortened function bodies

Hackerpilot:
Support for 2.096.0's new `-preview=shortenedMethods` syntax.

PR #454, opened 2022-02-01
allow in/out contract expressions in shortened methods

WebFreak001:
Fixes contract expressions with shortened methods introduced in the previous PR #439 

Might want to improve syntax errors with it, but this should work for now.


WebFreak001:
note: errors like `` `do` expected after InStatement or OutStatement`` seem to never be emitted. Unsure if this is because of this PR, though they aren't tested anyway. Might want to add error message tests in the future and improve some areas (like the fail file in fail_files/shortenedMethod.d as well)

PR #462, opened 2022-06-03
Add IfCondition to while (since DMD-FE 2.097.0)

WebFreak001:
to fix https://github.com/Pure-D/serve-d/issues/228


WebFreak001:
added more exhaustive tests and added support for `scope`, going to merge in

Issue #73, opened 2015-12-25
attr: decls doesn't attach attribute to the decls


adamdruppe:
```
$ dscanner --ast


@test void foo();


<?xml version="1.0"?>
<module>
<declaration>
<attribute>
<atAttribute>
<identifier>test</identifier>
</atAttribute>
</attribute>
<functionDeclaration line="1">
<name>foo</name><type pretty="void">
<type2>void</type2>
</type>
<parameters>
</parameters>
</functionDeclaration>
</declaration>
</module>
```

Yet:

```
$ dscanner --ast


@test: void foo();


<?xml version="1.0"?>
<module>
<declaration>
<attributeDeclaration>
<attribute>
<atAttribute>
<identifier>test</identifier>
</atAttribute>
</attribute>
</attributeDeclaration>
</declaration>
<declaration>
<functionDeclaration line="1">
<name>foo</name><type pretty="void">
<type2>void</type2>
</type>
<parameters>
</parameters>
</functionDeclaration>
</declaration>
</module>
```

Notice that the first one has the attribute inside the declaration, but the second one has it in a different declaration.

I think the way dmd does this is the colon starts a declaration list and the attribute is attached to everything in that list. Does the same with privacy keywords btw.

But, I think dmd also does that attachment as part of semantic analysis, so I'm not sure this is a bug strictly speaking... but I'd like to be sure.

PR #466, opened 2022-09-16
fix(parser): attributes are not always propagated/populated

### LINKED ISSUE #370: Attributes broken?
Either by using the visitor interface or serializing a whole module, we can see AtAttributes are being parsed (as are access specifiers) but they don't seem to be attached to FunctionDeclaration classes. I could be doing something wrong here, but as far as I can tell it is impossible to get attributes of functions without keeping your own track of the whole module separately 
```
/+dub.sdl:
dependency "libdparse" version="~>0.9"
+/
import dparse.ast;
import std.stdio;
import dparse.lexer;
class TestVisitor : ASTVisitor
{
    alias visit = ASTVisitor.visit;

    override void visit(const FunctionDeclaration decl)
    {
       
        decl.name.text.writeln;
        decl.attributes.writeln; //empty?
      	decl.memberFunctionAttributes.writeln;
        decl.accept(this);
        
        
    }
    override void visit(const AtAttribute decl)
    {
    	writeln("wtf");
    }
    
}

void main()
{
    import dparse.lexer;
    import dparse.parser : parseModule;
    import dparse.rollback_allocator : RollbackAllocator;
    import std.array : array;
    import std.string : representation;

    auto sourceCode = q{
        
        public @safe @wow pure void foo(int x)   
        {
        	pragma(msg, "dumb");
        }
        class test {
        	private const void functor() 
            {
            	
            }
        }
    }.dup;
    LexerConfig config;
    auto cache = StringCache(StringCache.defaultBucketCount);
    auto tokens = getTokensForParser(sourceCode.representation, config, &cache);
	
    RollbackAllocator rba;
    auto m = parseModule(tokens.array, "test.d", &rba);
    auto visitor = new TestVisitor();
    visitor.visit(m);
}
```

ljmf00:
Co-authored-by: Eyal Lotem <eyal@weka.io>
Signed-off-by: Luís Ferreira <contact@lsferreira.net>

---

Adapted from https://github.com/weka/libdparse/commit/caac5e48884992aebddf1f3bd051f5c128d505d6 to upstream. CC @EyalIO .

Fixes #370 and somewhat related to #73.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -2179,7 +2179,7 @@ class Parser
             }
             else if (isAuto == DecType.autoFun)
             {
-                mixin(nullCheck!`node.functionDeclaration = parseFunctionDeclaration(null, true)`);
+                mixin(nullCheck!`node.functionDeclaration = parseFunctionDeclaration(null, true, node.attributes)`);
                 node.tokens = tokens[startIndex .. index];
                 return node;
             }
@@ -2412,7 +2412,7 @@ class Parser
                 goToBookmark(b2);
                 if (savedComment && comment is null)
                     comment = savedComment;
-                node.functionDeclaration = parseFunctionDeclaration(t, false);
+                node.functionDeclaration = parseFunctionDeclaration(t, false, node.attributes);
             }
             else abandonBookmark(b2);
             if (!node.variableDeclaration && !node.functionDeclaration)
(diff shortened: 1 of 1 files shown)


ljmf00:
Still requires testing.

WebFreak001:
we have the XPath based testing framework setup for this, I think that's the easiest way to test for an AST modification like this.

PR #483, opened 2023-02-09
attach token location to error tokens

WebFreak001:
Makes invalid strings error at proper location instead of at 0,0

See confusion in https://forum.dlang.org/post/mcaldhecyexsolqoctsy@forum.dlang.org

### DIFF
--- a/src/dparse/lexer.d
+++ b/src/dparse/lexer.d
@@ -1208,8 +1208,7 @@ private pure nothrow @safe:
         {
             if (range.index >= range.bytes.length)
             {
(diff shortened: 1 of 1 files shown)


WebFreak001 (commented):
since checking for error tokens is done by comparing `== tok!""`, which only checks the .type property, this should be fully backwards compatible both for valid code and invalid code.

As a result of this PR incomplete strings, broken character literals, etc. now show proper error position when doing things such as `x = '';` inside `void main() {}` in the parser code.

PR #486, opened 2023-02-14
Fix #484, support multiple message parts in assert

### LINKED ISSUE #484: Lidbparse does not implement multiple messages for static assert
As per: https://dlang.org/changelog/2.102.0.html#dmd.static-assert

This is currently blocking phobos from using this feature.

WebFreak001:
Also modified the AST for regular asserts, for future proofing and easier implementation.

cc @RazvanN7

Introduces a deprecation in opEquals, it's fixed in #485

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -874,11 +874,21 @@ final class AssertArguments : BaseNode
 {
     override void accept(ASTVisitor visitor) const
     {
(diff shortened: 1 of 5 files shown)


RazvanN7:
Thanks!

RazvanN7:
Could also please push a new tag? That will unblock phobos PRs using this feature.

PR #489, opened 2023-03-22
make Statement, StatementNoCaseNoDefault and Declaration use SumType

WebFreak001:
old Declaration code used algebraic, which is more efficiently replaced with sumtype.

I replaced the mixin-generated properties with manual properties, since the code is very small for them anyway and helps them being shown in DCD

Since SumType works with well-defiend types, opEquals and visitIfNotNull was trivial to implement.

Saves around 0-3 MB of actual peak RAM in a ~170 MB test environment (could just be fluctuations how much it actually is, but it _will_ definitely include some RAM savings, since we no longer store a bunch of nulls for common types)

Easy backwards-compatible improvement I think.

Made setters backwards-incompatible on purpose, since we don't want accidental assignment of 2 different types on the same sumtype.

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -2158,7 +2158,7 @@ class Parser
             }
             if (currentIs(tok!":"))
             {
-                node.attributeDeclaration = parseAttributeDeclaration(attr);
+                node.setUnionType = parseAttributeDeclaration(attr);
                 mixin(nullCheck!`node.attributeDeclaration`);
                 ownArray(node.attributes, attributes);
                 node.tokens = tokens[startIndex .. index];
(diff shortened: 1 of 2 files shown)


src/dparse/parser.d
```diff
@@ -6354,7 +6354,7 @@ class Parser
      *     | $(LITERAL 'override')
      *     | $(LITERAL 'pure')
      *     | $(LITERAL 'ref')
-     *     | $(LITERAL '___gshared')
+     *     | $(LITERAL '__gshared')
```
WebFreak001, line 6357: maybe this was intended? It pops up here and there, need to investigate what this is about

Issue #409, opened 2020-05-20
PrimaryExpression and UnaryExpression are not decomposed


Jozef-r:
`PrimaryExpression` and `UnaryExpression` are not decomposed and this has a negative impact on the amount of memory used.

UnaryExpression should be

```d
class UnaryExpression : AstNode
{
    Expression exp;
}
```

and then

```d
class PrefixExpression : UnaryExpression {}
class PreIncrementExpression : PrefixExpression {}
class PreDecrementExpression : PrefixExpression {}
class AtExpression : PrefixExpression {}
class DereferenceExpression : PrefixExpression {}
class PostfixExpression :  UnaryExpression {}
class IndexExpression : PostfixExpression {}
class DotExpression : PostfixExpression {}
/**\ ETC \**/
```

For example `a.b.c` takes almost 3 kB of memory.

WebFreak001:
this is one of my primary things that concern me about the current API layout, let's try to get this in before a 1.0.0 release!

PR #490, opened 2023-03-23
separate ASTVisitor.visit for ExpressionNode

WebFreak001:
abstract classes should not have visitor functions, but rather only do dynamic dispatch. This becomes more important later with #409

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -92,11 +92,32 @@ enum DeclarationListStyle : ubyte
  */
 abstract class ASTVisitor
 {
-
-    /** */
+    deprecated("Don't use or override visit(ExpressionNode). For usage: dynamicDispatch(ExpressionNode) is equivalent; "
+        ~ "for overriding: you should probably override more specific cases. If you need to override to inject some "
+        ~ "before/after code for all cases, make sure to call `super.dynamicDispatch(n)` instead of `n.accept(this)`!")
     void visit(const ExpressionNode n)
     {
-        switch (typeMap[typeid(n)])
+        dynamicDispatch(n);
+    }
+
+    /**
+     * Looks at the runtime type of `n`, then calls the appropriate `visit`
+     * method at runtime.
+     *
+     * Rule of thumb: when the type is an abstract class, use `dynamicDispatch`,
+     * otherwise use `visit`.
+     *
+     * For templated calls:
+     * ---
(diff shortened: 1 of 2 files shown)


vushu (approved):
LGTM

WebFreak001:
the "Maximum resident set size" seems unlikely to have changed drastically with this PR (in fact, only stack usage might have slightly increased through the introduced additional function call for compatibility)

We probably need to investigate about a better approach for tracking peak memory usage in DCD. Maybe using valgrind massif?

WebFreak001:
merging with reduced coverage since the missing coverage is in a deprecated function that will be removed eventually. (and is still covered by DCD tests)

src/dparse/ast.d
```diff
@@ -92,11 +92,32 @@ enum DeclarationListStyle : ubyte
...
+     * ---
+     * static if (__traits(isAbstractClass, typeof(node)))
+     *   visitor.dynamicDispatch(node);
+     * else
+     *   visitor.visit(node);
+     * ---
+     */
+    void dynamicDispatch(const ExpressionNode n)
+    {
+        switch (typeMap.get(typeid(n), 0))
```
SixthDot, line 120: The following casts are not good as they could. There should be a bit cast based on the case:

    case 1: visit(*cast(AddExpression*) &n); break;
    /*etc*/
    case 48: visit(*cast(XorExpression*) &n); break;

So far there are still dynamic casts, i.e calls to runtime _d_obj_cast, although those calls always succeed.

WebFreak001: I don't think using bitcasts for class inheritance is safe, since the cast manipulates the pointer value, e.g.

> Casting a class object to an interface consists of adding the offset of the interface's corresponding vptr to the address of the base of the object. Casting an interface ptr back to the class type it came from involves getting the correct offset to subtract from it from the object.Interface entry at vtbl[0]. Adjustor thunks are created and pointers to them stored in the method entries in the vtbl[] in order to set the this pointer to the start of the object instance corresponding to the implementing method.

SixthDot: No interface involved here. `this` is still the same after both casts.

SixthDot: The compiler cant do it for you here as obviously the parameter is a base node.  I wont insist more but at least you are aware of the small issue now.

PR #491, opened 2023-03-23
Flatten unary expressions

WebFreak001:
Depends on #490 

Step 1/2 for #409 

RAM usage for one environment: 197 MB -> 130 MB (-33%)

CPU time about the same (ever so slightly lower, but could just be jitter)


rikkimax:
Something about this not being faster smells to me. Usually, when you decrease RAM usage significantly, it should be much faster.

But otherwise, less ram usage the better!

src/dparse/parser.d
```diff
@@ -3478,7 +3478,13 @@ class Parser
...
      *     ;)
      */
-    FunctionCallExpression parseFunctionCallExpression(UnaryExpression unary = null)
+    FunctionCallExpression parseFunctionCallExpression()
+    {
+        return parseFunctionCallExpression(null, false);
+    }
+
+    /// ditto
+    FunctionCallExpression parseFunctionCallExpression(UnaryExpressionNode unary, bool fromUnary = true)
```
vushu, line 3487: consider using enum instead of bool since it's more describtive and type strict
```
enum FromUnary {
      yes,no;
}
```

vushu (approved):
Looks good

PR #493, opened 2023-05-02
syntax changes from 2.104.0

WebFreak001:
supports for example `@true` or `@int` now


WebFreak001:
CI failure is due to `https://issues.dlang.org/show_bug.cgi?id=23874`

PR #494, opened 2023-05-03
Improve error recovery

WebFreak001:
improves changes from #403


WebFreak001:
DCD build failure is due to [issues.dlang.org/show_bug.cgi?id=23874](https://issues.dlang.org/show_bug.cgi?id=23874)

PR #508, opened 2023-10-28
add interpolated strings

WebFreak001:
based on [DIP1036e](https://github.com/dlang/dmd/pull/15715)

needs more tests still and actual usage in e.g. DCD to see how it behaves

TODO:
- token index and position should probably still stay relative to file start, not relative to istring start, so need to adjust that. Tthis will make it easier to find the cursor position inside (recursive) istrings.

### DIFF
--- a/src/dparse/lexer.d
+++ b/src/dparse/lexer.d
@@ -47,7 +47,7 @@ private immutable dynamicTokens = [
     "whitespace", "doubleLiteral", "floatLiteral", "idoubleLiteral",
     "ifloatLiteral", "intLiteral", "longLiteral", "realLiteral",
     "irealLiteral", "uintLiteral", "ulongLiteral", "characterLiteral",
-    "dstringLiteral", "stringLiteral", "wstringLiteral"
+    "dstringLiteral", "stringLiteral", "wstringLiteral", "istringLiteral",
 ];
 
 private immutable pseudoTokenHandlers = [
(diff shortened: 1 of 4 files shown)


src/dparse/lexer.d
```diff
@@ -2531,3 +2594,31 @@ void main() {
...
+
+        auto l = DLexer(srcAndDst, cf, &ca);
+        assert(l.front().type == tok!"istringLiteral", str(l.front.type));
+        assert(l.front().text == srcAndDst, l.front.text);
+        l.popFront();
+        assert(l.messages.empty);
+        assert(l.empty);
+    }
+
+    checkExactlyOneToken(`i"hello"`);
```
adamdruppe, line 2615: looks like your tests don't include iq{} and i\`\` so might want to add this (pushed up to the dmd thing just a couple days ago. i think that is all the string forms i intend to add at this time, though i imagine eventually someone will ask for the other delimited forms too lol)

adamdruppe:
tbh took a lil more code than i thought it would

but still... nice.

src/dparse/lexer.d
WebFreak001: I also added c/w/d literal suffix, I saw that was dismissed in the PR?

adamdruppe: Yeah, it lexes and parses them, but then issues a semantic error.

WebFreak001:
closing in favor of #510

PR #510, opened 2023-10-29
interpolated strings, v2

WebFreak001:
better implementation, doing all the tokenizing at the start, an interpolated string being multiple tokens

requires no further changes in DCD to support with auto-completion, probably also going to just work with D-Scanner and serve-d

supersedes #508 

implementation for [DIP1036e](https://github.com/dlang/dmd/pull/15715)

### DIFF
--- a/src/dparse/ast.d
+++ b/src/dparse/ast.d
@@ -16,10 +16,10 @@
 module dparse.ast;
 
 import dparse.lexer;
-import std.traits;
 import std.algorithm;
 import std.array;
 import std.string;
+import std.traits;
 
 private immutable uint[TypeInfo] typeMap;
 
@@ -1425,8 +1425,8 @@ final class Declaration : BaseNode
         }
     }
 
-    private import std.variant:Algebraic;
-    private import std.typetuple:TypeTuple;
+    import std.typetuple : TypeTuple;
+    import std.variant : Algebraic;
 
     alias DeclarationTypes = TypeTuple!(AliasDeclaration, AliasAssign, AliasThisDeclaration,
         AnonymousEnumDeclaration, AttributeDeclaration,
@@ -3483,9 +3483,10 @@ final class TemplateSingleArgument : BaseNode
 {
     override void accept(ASTVisitor visitor) const
     {
-        mixin (visitIfNotNull!(token));
+        mixin (visitIfNotNull!(token, istring));
     }
     /** */ Token token;
+    /** */ InterpolatedString istring;
     mixin OpEquals;
 }
 
@@ -4092,7 +4093,7 @@ unittest //#365 : used to segfault
 unittest // issue #398: Support extern(C++, <string expressions...>)
 {
     import dparse.lexer : LexerConfig;
-    import dparse.parser : ParserConfig, parseModule;
+    import dparse.parser : parseModule, ParserConfig;
     import dparse.rollback_allocator : RollbackAllocator;
 
     RollbackAllocator ra;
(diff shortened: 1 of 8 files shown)


schveiguy:
I'm all for merging this. I don't really have a good feel for what is the proper way to do these things, but we definitely need something. As noted from the linked bug reports, people are finding the lack of istring support troublesome, and it will only get worse. I think it is better to have a buggy or substandard implementation than no attempt.

Needs a rebase though...

schveiguy (approved):
Caveat: I'm not a compiler dev, so this is an approval of the improvement, not necessarily the code. Nothing stands out as bad to me.

rikkimax:
Same idea as in the compiler.

The lexer has a mini parser that consumes tokens and spits out others.

It at least lines up.

unleashy:
Not sure this is still being worked on, but with this branch, iq{...} with line breaks in them are syntax errors, even though dmd accepts them just fine (as you would expect); q{...} with linebreaks parses fine.

adamdruppe:
i gotta get this into adrdox and the code looks ok to me, i gotta do adaptation to fit in my code but i think i can use most of it so ill let you know how it goes (though i avoid using new features in the libraries so it isn't going to be a very comprehensive test, still it is some)

adamdruppe:
can confirm the `iq{ ... }` ones are broken. I get a differnet error, but my branch divered back in 2016 so that's not gonna be too helpful (tbh im amazed most the patch applied automatically in my branch! the parser and formatter didn't, but those aren't important to adrdox anyway. and the lexer patch applied successfully and is mostly working).

Looks like the first { is not seen by the lexIstringPlain function. In `lexInterpolatedString`, it detects the `case 'q'` then does a `range.popFront` inside that switch. There's a second range.popFront below that consumes the `{` prematurely.

I suggest deleting the `range.popFront` inside of `case 'q'`.

Then in lexIstringPlain:

```
            case '}':
                if (istringStack[$ - 1].type == IstringState.Type.tokenString)
                {
                    assert(istringStack[$ - 1].braces > 0);
                    istringStack[$ - 1].braces--;
                   // add this to recognize the end of the line
                    if(istringStack[$ - 1].braces == 0)
                        break Loop;
                }
                goto default;
```


but this fails:

```
        auto a = iq{ "}" hi };
```

so i don't think this approach is correct at all really, it sh ould be running a child lexer inside the q{} thing.

adamdruppe:
Yeah, I think the best approach is to copy/paste from `lexTokenString` and inject checks for `tok!"$"` to do the escaping behavior instead of reading the bytes. Then the lexer will handle the quoted string and char literals correctly. That's pretty much what dmd does and it would work better here too.

tho tbh for now adrdox is just gonna leave it as a `// FIXME` cuz realistically this is better than nothing and im out of time to mess with it tonight.

LightBender:
Bump. What would it take to get this over the finish line? We have a backup of PR's waiting on this.

WebFreak001:
missing: tests & fixes for the iq{} strings

LightBender:
> missing: tests & fixes for the iq{} strings, like mentioned in the comments by others

Any chance we can get this resolved/merged during/prior to BeerConf this month?

schveiguy:
This really needs to be figured out...

pbackus:
This is blocking https://github.com/dlang/phobos/pull/10652. Can we maybe merge as-is and do `iq{...}` in a follow-up PR? Partial support for IES is still better than nothing.

LightBender:
@pbackus I'm OK with doing a partial merge, but this particular branch has merge conflicts.

WebFreak001:
support should be there now for all cases, please throw test cases at me

LightBender (approved):

pbackus:
For test cases, these may be of interest: https://github.com/dlang/dmd/blob/master/compiler/test/runnable/interpolatedexpressionsequence.d

WebFreak001:
image

D-Scanner and DCD support comes for free with this PR

pbackus:
#524 should fix the CI failure

PR #512, opened 2023-12-29
Fix shortened FunctionBody and comment formatting

zyebytevt:
Nothing major, but the comment formatting produced invalid code since the three slashes at the beginning weren't written. Additionally, a shortened function body (`=>`) now also appears in formatting.

### DIFF
--- a/src/dparse/formatter.d
+++ b/src/dparse/formatter.d
@@ -1178,7 +1178,7 @@ class Formatter(Sink)
             if (member.comment.length)
             {
                 space();
-                put(member.comment);
+                putComment(member.comment);
             }
         }
         endBlock();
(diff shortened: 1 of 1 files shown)


WebFreak001:
add a test in the tests folder to show which issue this actually fixes

WebFreak001 (changes requested):
needs test

zyebytevt:
Where should I add tests for this? My changes solely concern the `Formatter`, and if I'm not mistaken there appear to be no prior tests for it at all.

WebFreak001:
ah sorry I thought this was in dfmt. In `libdparse` at the bottom there is a large unittest block where a bunch of code snippets are tested.

You will probably want something like
```d
testFormatNode!(FunctionDeclaration)(q{void someFunction() => whateverWasFixed()});
```

if you only provide a single string it is parsed and outputted again and checked that it matches the input exactly.

zyebytevt:
Alright, that should hopefully do it. Had to keep in mind various empty lines for the third check that the formatter produces, but that seems to be an issue (or intended) from before.

WebFreak001 (approved):
not ideal, but better than before and not really that essential to be perfect atm

PR #514, opened 2024-03-20
Use @trusted qualifier cast

ntrel:
Needed for https://github.com/dlang/dmd/pull/16315.


WebFreak001 (approved):
test failure is due to DMD build with -profileGC regressing / unable to be runnable which has been the case for a few months now

PR #518, opened 2024-04-09
fix grammar documentation generation

ricardaxel:
This issue comes from ddox tokenizer that seems to have a bug when escaped quotes appear in [token strings](https://dlang.org/spec/lex.html#token_strings).

I guess a long-term solution would be to report and fix the bug directly in ddox


WebFreak001 (approved):
good find

PR #530, opened 2026-02-28
Allow keywords as identifiers in extern(C) blocks for better C interopability

bangbangsheshotmedown:
-Hf ouputs all symbols inside a `extern(C) {}` for ImportC

libdparse struggle with that if one of the symbol's identifier is a reserved keyword

This PR introduces a new config option for the Parser, `bool importC`, if set to true it'll relax the rule to allow keywords to be used as identifier, so it can proceed in parsing the whole `extern(C) {}` block properly

### DIFF
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -5532,7 +5532,7 @@ class Parser
 
         ownArray(node.parameterAttributes, parameterAttributes);
         mixin(parseNodeQ!(`node.type`, `Type`));
-        if (currentIs(tok!"identifier"))
+        if (currentIs(tok!"identifier") || (isInExternC() && isKeyword(current.type)))
         {
             node.name = advance();
             if (currentIs(tok!"..."))
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -2303,7 +2303,7 @@ class Parser
             warn("Empty declaration");
             advance();
             break;
-    case tok!"{":
+        case tok!"{":
         if (node.attributes.empty)
         {
             error("declaration expected instead of `{`");
--- a/src/dparse/parser.d
+++ b/src/dparse/parser.d
@@ -118,6 +118,16 @@ Module parseModule(F)(const(Token)[] tokens, string fileName, RollbackAllocator*
  */
 class Parser
 {
+    /// Is it a interface file for a C source code
+    bool importC;
+    /// Depth of extern(C) blocks we're currently inside
+    int externCDepth;
+
+    bool isInExternC() const pure nothrow @safe @nogc
+    {
+        return importC || externCDepth > 0;
+    }
+
     /**
      * Parses an AddExpression.
      *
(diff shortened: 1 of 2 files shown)


src/dparse/parser.d
```diff
@@ -2293,27 +2303,44 @@ class Parser
             warn("Empty declaration");
             advance();
             break;
-        case tok!"{":
-            if (node.attributes.empty)
+    case tok!"{":
```
WebFreak001: what's up with the new indentation? (everything is detended one tab now)

bangbangsheshotmedown: copy pasta failure

src/dparse/parser.d
```diff
@@ -5495,7 +5532,7 @@ class Parser
 
         ownArray(node.parameterAttributes, parameterAttributes);
         mixin(parseNodeQ!(`node.type`, `Type`));
-        if (currentIs(tok!"identifier"))
+        if (currentIs(tok!"identifier") || (isInExternC() && isKeyword(current.type)))
         {
             node.name = advance();
```
WebFreak001, line 5537: we need a conversion from keyword token to identifier token here, so that `node.name.text` is set (otherwise it's unset for keywords iirc)

src/dparse/parser.d
```diff
@@ -2294,26 +2304,43 @@ class Parser
...
+            if (!declarations.put(parseDeclaration(strict, false, inTemplateDeclaration)))
             {
-                auto c = allocator.setCheckpoint();
-                if (!declarations.put(parseDeclaration(strict, false, inTemplateDeclaration)))
-                {
-                    allocator.rollback(c);
-                    return null;
-                }
+                allocator.rollback(c);
+                if (isExternC)
```
WebFreak001, line 2334: use `scope (exit)` instead of duplicating the `if (isExternC) externCDepth--;` at both exit points

src/dparse/parser.d
```diff
@@ -2481,7 +2508,7 @@ class Parser
         foreach (B; BasicTypes) { case B: }
         type:
             Type t = parseType();
-            if (t is null || !currentIsOneOf(tok!"identifier", tok!":"))
+            if (t is null || !(currentIs(tok!"identifier") || (isInExternC() && isKeyword(current.type)) || currentIs(tok!":")))
```
WebFreak001, line 2511: we should probably not allow all keywords, only what is allowed in C context (e.g. we need a new helper function instead of `isKeyword` something like `isCIdentifier()`

src/dparse/parser.d
```diff
@@ -2654,7 +2681,17 @@ class Parser
...
+            else
+            {
+                error("Expected identifier instead of " ~ (index < tokens.length
+                    ? (tokens[index].text is null ? "`" ~ str(tokens[index].type) ~ "`"
+                       : str(tokens[index].type) ~ " `" ~ tokens[index].text ~ "`")
+                    : "EOF"));
+                return null;
+            }
             mixin (nullCheck!`id`);
             node.name = *id;
```
WebFreak001, line 2696: we need a conversion from keyword token to identifier token here, so that node.name.text is set (otherwise it's unset for keywords iirc)

test/tester.d
```diff
@@ -313,34 +313,63 @@ void testArbitraryASTs()
...
+    {
+        parser = new Parser();
+        parser.messageDelegate = &msgDelegate;
+        parser.allocator = &rba;
+        parser.importC = false;
+        parser.tokens = getTokensForParser("extern(C) { int normal = 1; int true = 1; }", config, &cache);
+        assert(parser.parseAttribute() !is null);
+        auto b = parser.parseBlockStatement();
+        assert(b !is null);
+        assert(b.declarationsAndStatements.declarationsAndStatements.length == 1);
```
WebFreak001, line 358: please add an `assert` for the `errors` array here

WebFreak001 (commented):
thanks for the PR, I requested some changes that improve code quality and some that are needed for consumers to be able to properly work with the new behavior

PR #531, opened 2026-04-07
Fix #526, support `int*[0].init`

### LINKED ISSUE #526: Support `int*[0].init`
Since 2.102.2 (found with run.dlang.org), the following is valid:
```d
	enum E : int*[0] { a = int*[0].init }
```
dfix with libdparse [updated](https://github.com/dlang-community/dfix/pull/61) to 0.25.1 causes these errors for the above line, on the `*` token:
```
/home/nick/dr/src/core/internal/traits.d(652:35)[error]: Primary expression expected
/home/nick/dr/src/core/internal/traits.d(652:39)[error]: `,` or `}` expected (found token `.`)
```

0-v-0:


src/dparse/parser.d
WebFreak001:
```suggestion
     *     | $(RULE builtinType) $(RULE typeSuffix)* $(RULE arguments)
```

WebFreak001:
```suggestion
     *     | $(RULE builtinType) $(RULE typeSuffix)* $(LITERAL '.') $(LITERAL Identifier)
```

WebFreak001:
```suggestion
     *     | $(RULE typeConstructor) $(LITERAL '$(LPAREN)') $(RULE type) $(LITERAL '$(RPAREN)') $(LITERAL '.') $(LITERAL Identifier)
```

src/dparse/parser.d
```diff
@@ -5923,18 +5923,31 @@ class Parser
                 break;
             }
         foreach (B; BasicTypes) { case B: }
-            node.basicType = advance();
+        {
+            auto bookmark = setBookmark();
+            auto c = allocator.setCheckpoint();
+            if (auto t = parseType())
```
WebFreak001: the condition always returns true (must be Type -> Type2 -> BuiltinType at this point due to the current token being one of BasicTypes and then the parseType does not require any further things)

Also removes the need for bookmark/checkpoint

Additionally for backwards compatibility add a getter function in ast.d in PrimaryExpression, which returns `type?.type2?.builtinType ?? IdType.init`

WebFreak001 (changes requested):

WebFreak001 (approved):

PR #532, opened 2026-04-08
Fix #523, support align(default)

### LINKED ISSUE #523: Support `align(default)` once upstream does
Upstream D has plans to support `align(default)` in the future: <https://github.com/dlang/dmd/issues/20587> <https://github.com/dlang/dmd/pull/20589>

At the moment, I’m getting the following errors with serve-d on the upstream test file:

```d
struct S
{
    align(1)
    {
        short x1;
        int y1;
        long z1;

        align(default) /* Primary expression expected */
        { /* declaration expected instead of `{` */
            short x;
            int y;
            long z;
        }
    }
}
```

0-v-0:


src/dparse/parser.d
```diff
@@ -312,7 +312,10 @@ class Parser
         if (currentIs(tok!"("))
         {
             mixin(tokenCheck!"(");
-            mixin(parseNodeQ!("node.assignExpression", "AssignExpression"));
+            if (currentIs(tok!"default"))
+                advance();
```
WebFreak001, line 318: this is implicitly stored in the AST now (due to it having the `tokens` slice), but it's missing a boolean getter (e.g. `bool isExplicitDefault()`) inside AlignAttribute

The `formatter.d` code also needs to be adjusted to re-emit it in case it was explicit default.

src/dparse/parser.d
WebFreak001: Missing documentation update


```suggestion
     *     $(LITERAL 'align')
     *     $(LITERAL 'align') $(LITERAL '$(LPAREN)') $(RULE assignExpression) $(LITERAL '$(RPAREN)')
     *     $(LITERAL 'align') $(LITERAL '$(LPAREN)') $(LITERAL 'default') $(LITERAL '$(RPAREN)')
```

WebFreak001 (changes requested):

WebFreak001 (approved):

